PR #7, opened 2022-08-10
Add missing patch perms

abays:
Add missing patch permissions for several resource types


abays:
There are other missing permissions.  Switching to draft to add

abays:
The errors in the linting are due to new version of lib-common with condition changes somehow being used.  That version is NOT used in this version of the Cinder operator, however, and these errors do not occur when you use the operator

dprince (approved):

PR #9, opened 2022-08-31
Add external Ceph cluster support

fmount:
This patch introduces the API extension to support an external Ceph cluster in
the cinder operator. If Ceph is present (`CephMonHosts`, `CephClusterFSID` and
`CephClienKey` passed as input), the rbd driver is configured and can be
enabled through the `customServiceConfig` volume API key. The common-API
structure has been extended and now includes the `CephBackend` parameters used
to provide all the information required to setup the `rbd` backend when the
config file is created. The `CephBackend` belongs to the common-API since it's
common to have an external Ceph cluster serving multiple instances/workloads.

The Ceph Backend also supports the following configurable parameters:

- `CephUser`: the rbd user which is reflected within the keyring template and
   cinder.conf

- `CephPools`: map of service/`CephPoolSpec` that includes the `ExtraPools`
   definition we have in `TripleO`

Those parameters, used as input, are processed to properly build the Ceph
related data structures, mounted later in process by pods (mostly
`cinderVolumes`) that are supposed to interact with the external Ceph cluster.


fmount:
@dprince @abays This is pretty much the same thing we did in the glance-operator. Due to the nature of the component, at some point I imagine the two implementations taking two different paths, but atm, being able to create a volume using an external Ceph backend remains the priority.

fmount:
@gfidente @fultonj FYI

controllers/cinder_controller.go
```diff
@@ -637,6 +637,19 @@ func (r *CinderReconciler) generateServiceConfigMaps(
...
+	cephClient := cinder.GetCephBackend(instance)
+
+	/** If CephBackend info are passed, populate the required templateParameters
+	to make sure Cinder is able to interact with an external Ceph cluster
+	using the Client Key provisioned on Ceph
+	**/
+	if cephClient {
+		templateParameters["CephClusterFSID"] = instance.Spec.CephBackend.CephClusterFSID
+		templateParameters["CephClusterMonHosts"] = instance.Spec.CephBackend.CephClusterMonHosts
+		templateParameters["CephClientKey"] = instance.Spec.CephBackend.CephClientKey
```
abays, line 650: If `cephClient` is `false` and these parameters are then not set in the `templateParameters` map, does the template processing work?  I don't recall if the templating mechanism gracefully ignores `nil` template variable references or throws an exception.

abays: Answered my own question: seems to be fine.  Tried the new Glance operator and I see that it just dumps `<no value>` in for missing keys:

```
$ oc rsh glance-6dff4b4dd7-kb887
sh-5.1# cat /var/lib/config-data/merged/ceph.conf 
[global]
fsid = <no value>
mon_host = <no value>
```

fmount: and since we control `enabled_backends`, that section and those files are totally ignored if no ceph.

templates/cinder/config/cinder.conf
```diff
@@ -79,3 +79,12 @@ lock_path = /var/lib/cinder/tmp
...
 #nfs_shares_config=/etc/cinder/nfs_shares
+
+[ceph]
+volume_backend_name=ceph
+volume_driver=cinder.volume.drivers.rbd.RBDDriver
+rbd_ceph_conf=/etc/ceph/ceph.conf
+rbd_user=openstack
+rbd_pool=volumes
+rbd_flatten_volume_from_snapshot=False
+rbd_secret_uuid={{ .CephClusterFSID }}
```
abays, line 90: Please forgive my ignorance here, but is there anything else in this section that we might want to parameterize in addition to `rbd_secret_uuid`?  I'm not familiar with Ceph to any significant extent, but I'm curious if `rbd_user`, `rbd_pool`, etc, could vary between different external Ceph deployments.

fultonj: Yes, most definitely. We'll be submitting follow up PRs to parameterize it all. Same for glance.

So do we do that in one PR or in a few follow ups?

fmount: > Please forgive my ignorance here, but is there anything else in this section that we might want to parameterize in addition to `rbd_secret_uuid`? I'm not familiar with Ceph to any significant extent, but I'm curious if `rbd_user`, `rbd_pool`, etc, could vary between different external Ceph deployments.

Yeah, I'm updating this PR (since it's already here) to add the more parameters. @fultonj feel free to make the same thing for Glance.

api/v1beta1/common_types.go
```diff
@@ -48,3 +48,35 @@ type CinderDebug struct {
...
+	CephClientKey string `json:"cephClientKey,omitempty"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="CephUser"
+	// CephUser set the Ceph cluster pool used by Cinder
+	CephUser string `json:"cephUser,omitempty"`
+	// +kubebuilder:validation:Optional
+	// CephPools - Map of chosen names to spec definitions for the Ceph cluster
+	// pools used by Cinder
+	CephPools map[string]CephPoolSpec `json:"cephPools,omitempty"`
+}
```
abays, line 74: Just curious as to whether these should be required if the user has chosen to define/include a `CephBackend` in the `CinderSpec`.  If the user has included a `CephBackend` in the main spec, is it valid to set some of these and not set others?  Or did you perhaps make it this way because the k8s API validator was complaining without everything being optional?  I think I recall seeing k8s validation not allow unset sub-fields even if the containing field itself was optional and the user had not included it in the spec.

api/v1beta1/common_types.go
```diff
@@ -48,3 +48,35 @@ type CinderDebug struct {
...
+
+// CephPoolSpec defines the Ceph pool Spec parameters
+type CephPoolSpec struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="volumes"
+	// CephPoolName defines the name of the pool
+	CephPoolName string `json:"name,omitempty"`
+}
```
abays, line 82: Is the idea of using a struct here to support future parameters beyond just a name?  Because I guess otherwise the `CephPools` field in `CinderCephBackend` could just be a `map[string]string`.

pkg/cinder/ceph_backend.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// Build the resulting caps from the _ordered_ array applying the template
+	for _, pool := range plist {
+
+		osdCaps += func(p string) string {
+			var d bytes.Buffer
+			t.Execute(&d, T{p})
+			return (d.String())
+		}(pool)
+	}
```
abays, line 111: Just curious as to why a template was needed here.  `fmt.Sprintf()` seems like it might work and would be simpler, but perhaps I am missing something.

api/v1beta1/common_types.go
fmount: Yeah, thanks, this is a good point.
We require at least to set `CephClusterFSID`,  `CephClusterMonHosts` and the `CephClientKey`. These are the minimum amount of info we need to start the Ceph backend.
In general, if `CephUser` and `CephPools` are not specified, we have defaults in place, which means  we can keep them as "Optional".
Now, being that said,  I think we can change `kubebuilder:validation:Optional` to `kubebuilder:validation:Required`, what do you think?

api/v1beta1/common_types.go
fmount: Yes, that's the reason.
In future we can have some overrides in place that can influence the way the service is configured (e.g. tiering depending on the `crush rules` set on the ceph cluster).
It's a good idea having this kind of struct here to help working on small changes (easier to review) in the follow up patches.

abays: Ack, sounds good

pkg/cinder/ceph_backend.go
fmount: mmm I guess it's just a different approach to achieve the same result :D

api/v1beta1/common_types.go
abays: Sure, that seems reasonable.  Also, when we add that annotation, we should remove `omitempty`.  It's weird that there are sort of two ways to declare "optional", but it is what it is I suppose.  

I do wonder though if we'll run into the validation concern I mentioned (about required sub-fields in an optional containing field).  I guess we'll find out. :)

fmount: Yeah, we should remove both `omitempty` and the `default`, otherwise from a quick test I see the k8s API validator set the that value to the field even if not provided!
Removing both `omitempty` and the `+kubebuilder:default=<Value>` block the user during the object creation.

E.g.
```
> oc create -f out/openstack/cinder/cr/cinder_v1beta1_cinder.yaml
The Cinder "cinder" is invalid: spec.cephBackend.cephClientKey: Required value
```

abays (approved):
/lgtm

pkg/cinder/ceph_backend.go
fmount: Ok, simplifying this part adding fmt.Sprintf

fmount: Done.

PR #11, opened 2022-09-06
Rework CephBackend and rely to the lib-common Ceph Module

fmount:
This patch reworks the way cinder-operator defines the CephBackend
and interacts with an external Ceph cluster.
Since most of the OSP related operators are going to share the same
Ceph data structure and functions, we can leave here the logic that
is supposed to configure the service as long as the CephBackend is
defined, importing the required utilities from lib-common.


fmount:
Still a draft that can be rebased/updated once lib-common includes the Ceph module.

controllers/cinder_controller.go
```diff
@@ -645,15 +646,15 @@ func (r *CinderReconciler) generateServiceConfigMaps(
...
 	if cephClient {
-		templateParameters["CephClusterFSID"] = instance.Spec.CephBackend.CephClusterFSID
-		templateParameters["CephClusterMonHosts"] = instance.Spec.CephBackend.CephClusterMonHosts
-		templateParameters["CephClientKey"] = instance.Spec.CephBackend.CephClientKey
+		templateParameters["ClusterFSID"] = instance.Spec.CephBackend.ClusterFSID
+		templateParameters["ClusterMonHosts"] = instance.Spec.CephBackend.ClusterMonHosts
+		templateParameters["ClientKey"] = instance.Spec.CephBackend.ClientKey
 		// The pool we write in cinder.conf
-		templateParameters["CephPool"] = cinder.GetCephCinderPool(instance)
+		templateParameters["Pool"], _ = ceph.GetPool(instance.Spec.CephBackend.Pools, instance.Name)
```
abays, line 653: If the pool is not yet found in `instance.Spec.CephBackend.Pools`, it looks like it uses the second parameter to determine the default pool name: https://github.com/openstack-k8s-operators/lib-common/blob/e774739df18ae3a5be4207437646cd0fb29b5bb9/modules/storage/ceph/funcs.go#L28-L44.  But in that case, if the user named the instance something other than, say, `cinder`, the call to `ceph.GetPool` would return an error.  Perhaps we should check for an error here and also maybe pass something other that `instance.Name` as the service name?

fmount: Yeah, that's true, I thought the operator hadn't the ability to customize this name.
As you can see in [0], for cinder operator we expect to see "cinder" as generic name of the service: I'm wondering if extra logic is just useless complexity and we can just change this line to:

```
templateParameters["Pool"], _ = ceph.GetPool(instance.Spec.CephBackend.Pools, "cinder")
```
In this case we ensure to return "volumes" if the cinder pool is not specified in the CRD.

[0]  https://github.com/openstack-k8s-operators/lib-common/blob/e774739df18ae3a5be4207437646cd0fb29b5bb9/modules/storage/ceph/funcs.go#L28-L44

fmount: Restored the previous get where "cinder" is specified as a string. In future, service names default can be provided by the ceph module in lib-common, and at that point we can update the operators accordingly.

abays: +1

abays:
+1 from me, but I'll wait to approve until John or Dan takes a look

fultonj (approved):
/lgtm thanks

abays (approved):
/lgtm

Akrog:
Did anybody run this thing before merging?
Because it's broken, the cinder template has not been updated. :-(

PR #12, opened 2022-09-09
Fix Cinder services hanging

Akrog:
Cinder services as deployed by the operator just hang and will enter an unending loop of kill and restart due to the Liveness probes.

What we see in the container logs differ from the cinder-api to the other containers.

In the cinder-api we just see that it stops responding to requests, and on the other services we see this exception:

```
  Traceback (most recent call last):
    File "/usr/lib/python3.9/site-packages/eventlet/hubs/hub.py", line 476, in fire_timers
      timer()
    File "/usr/lib/python3.9/site-packages/eventlet/hubs/timer.py", line 59, in __call__
      cb(*args, **kw)
    File "/usr/lib/python3.9/site-packages/eventlet/semaphore.py", line 152, in _do_acquire
      waiter.switch()
  greenlet.error: cannot switch to a different thread
```

In both cases the issue is the same, there is some logging happening on a native thread and this is creating problems for eventlet, to the point where it hangs.

This is a known bug in eventlet [1], one which I recently fixed in Oslo-Log [2].

Since this is not fixed in all OpenStack releases, certainly not the one this operator is currently using, we need to be careful with what we actually enable for logging.

The logging we currently have enables debugging for EVERYTHING (rabbit, sqlalchemy, oslo libraries, etc.), regardless of what we set in the `debug` option and `default_log_levels` in `cinder.conf`.

This logging override is done via the `logging.conf` file and creates the problem of the native thread logging.

Using the `logging.conf` file diverges from the approach we want for the Cinder Operator, where we try to make the configuration of the Cinder services with the operator be as close as possible to a manual Cinder service configuration.

This patch removes the usage of the `logging.conf` file by the operator and uses the `cinder.conf` template to set the right logging configuration options.

We set `log_file = /dev/stdout` in `cinder.conf` instead of the usual `log_file =` because then Cinder services would log to `stderr` and make `httpd` on the cinder-api container treat all Cinder-API logs as errors, prepending additional information to every single cinder log message, like this:

```
  Thu Sep 08 08:21:36.404638 2022] [wsgi:error] [pid 15:tid 69] (sqlalchemy.orm.mapper.Mapper): 2022-09-08 08:21:36,404 INFO
```

References to the `logging.conf` file have been removed from CRD descriptions and other code locations.

[1]: https://github.com/eventlet/eventlet/issues/432
[2]: https://review.opendev.org/c/openstack/oslo.log/+/852443


abays (approved):
/lgtm

fultonj:
Logging to stdout instead of the stderr makes sense to me. Especially since otherwise cinder-api logs are treated as errors.

PR #13, opened 2022-09-09
Cleanup httpd.conf

Akrog:
The current WSGI VirtualHost for Cinder API is a big dirty.

We have the same VirtualHost configuration in 2 files, `10-cinder_wsgi.conf` and `httpd.conf` which makes it hard to know which one is actually being used.

The reason why it's duplicated is because we currently have the inclusion of the whole `conf.d` directory where `10-cinder_wsgi.conf` resides commented to disable SSL.

This patch cleans this up a bit and makes sure that we only have the VirtualHost configured in one place, `10-cinder_wsgi.conf` and we either include the configuration files from `conf.d` file or explicitly the Cinder wsgi configuration.

This removes confusion.

### DIFF
--- a/templates/cinder/config/httpd.conf
+++ b/templates/cinder/config/httpd.conf
@@ -12,8 +12,6 @@ Listen 8776
 TypesConfig /etc/mime.types
 
 Include conf.modules.d/*.conf
-# XXX: To disable SSL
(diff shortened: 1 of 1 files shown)


abays (approved):
/lgtm

PR #14, opened 2022-09-09
Fix ConfigMap generation ignores failures

Akrog:
Failures during the rendering and creation of the Config Maps are being ignored.

This patch changes this behavior and make generation failures prevent the operator from advancing in the process of deploying Cinder, after all without the Config Maps the services will not run as expected or not run at all.


abays (approved):
/lgtm

PR #15, opened 2022-09-09
Fix debug mode

Akrog:
The operator will fail to deploy Cinder volume, backup, and scheduler containers when the Service is set to debug mode:

  debug:
    service: true

This is because the startup probe will be missing an execution command.

In this patch we simply add the same dummy check for the startup probe as we do in the liveness and readiness probes and make a small change in the different probe exec setting to use the same instance for all probes instead of creating different instances with the same contents for each probe.


abays (approved):
/lgtm

PR #16, opened 2022-09-09
Log the Cinder CR instance name

Akrog:
Most of the logging in the Cinder operator don't show the name of the Cinder CR instance that is being worked on.

This patch simply adds the name to all the logs.


abays (approved):
/lgtm

We use these logging messages in other operators as well, so we should make a note to improve them there too

PR #17, opened 2022-09-09
Get Service by name using keystone

fmount:
As we've done in [0], it's a good idea using keystone to get the existing service spec instead
of explicitly building an object. This patch simplifies the logic in the `reconcileDelete` flow.

[0] https://github.com/openstack-k8s-operators/glance-operator/pull/31

### DIFF
--- a/controllers/cinderapi_controller.go
+++ b/controllers/cinderapi_controller.go
@@ -253,20 +253,19 @@ func (r *CinderAPIReconciler) reconcileDelete(ctx context.Context, instance *cin
 	// It's possible to get here before the endpoints have been set in the status, so check for this
 	if instance.Status.APIEndpoints != nil {
 		for _, ksSvc := range keystoneServices {
-			ksSvcSpec := keystonev1.KeystoneServiceSpec{
-				ServiceType:        ksSvc["type"],
-				ServiceName:        ksSvc["name"],
-				ServiceDescription: ksSvc["desc"],
-				Enabled:            true,
-				APIEndpoints:       instance.Status.APIEndpoints[ksSvc["name"]],
-				ServiceUser:        instance.Spec.ServiceUser,
-				Secret:             instance.Spec.Secret,
-				PasswordSelector:   instance.Spec.PasswordSelectors.Service,
+
+			ks, err := keystone.GetKeystoneServiceWithName(ctx, helper, ksSvc["name"], instance.Namespace)
+
+			if err != nil {
+				if k8s_errors.IsNotFound(err) {
+					return ctrl.Result{}, nil
+				}
+				return ctrl.Result{}, err
 			}
 
-			ksSvcObj := keystone.NewKeystoneService(ksSvcSpec, instance.Namespace, map[string]string{}, 10)
+			ksSvcObj := keystone.NewKeystoneService(ks.Spec, instance.Namespace, map[string]string{}, 10)
+			err = ksSvcObj.Delete(ctx, helper)
 
-			err := ksSvcObj.Delete(ctx, helper)
 			if err != nil {
 				return ctrl.Result{}, err
 			}


controllers/cinderapi_controller.go
```diff
@@ -253,20 +253,11 @@ func (r *CinderAPIReconciler) reconcileDelete(ctx context.Context, instance *cin
...
-				Enabled:            true,
-				APIEndpoints:       instance.Status.APIEndpoints[ksSvc["name"]],
-				ServiceUser:        instance.Spec.ServiceUser,
-				Secret:             instance.Spec.Secret,
-				PasswordSelector:   instance.Spec.PasswordSelectors.Service,
-			}
-
-			ksSvcObj := keystone.NewKeystoneService(ksSvcSpec, instance.Namespace, map[string]string{}, 10)
 
+			ks, _ := keystone.GetKeystoneServiceWithName(ctx, helper, ksSvc["name"], instance.Namespace)
```
Akrog, line 257: This has the downside of having additional requests, because we are getting the `KeystoneService` CR instead of building the spec ourselves, but it's cleaner.

We should check that the service was actually found and we didn't get an error (renaming `_` to `err` and checking if it's not `nil`), if it's an error and it's not found (`if k8s_errors.IsNotFound(err) {`) then we can assume it was deleted, otherwise we must fail.

controllers/cinderapi_controller.go
```diff
@@ -253,20 +253,11 @@ func (r *CinderAPIReconciler) reconcileDelete(ctx context.Context, instance *cin
...
+			ksSvcObj := keystone.NewKeystoneService(ks.Spec, instance.Namespace, map[string]string{}, 10)
```
abays, line 258: I know we're already using this pattern elsewhere (as you noted), but there are two things I was just pondering (and I apologize for not asking these earlier in the similar Glance PR):

1. What happens if the `ks` object comes back as `nil`?
2. Also, is it possible that `ks.Spec` is empty even if `ks` is non-`nil`?

Might be something that's only possible if the user is messing around at the CLI and manually deletes the `KeystoneService` CR.

Akrog: I think if we check for the returned `err` we won't ever get `nil`.  And we could also have a connection issue to etcd and have an error.

abays: That sounds like a good approach to me re: check for `err`

controllers/cinderapi_controller.go
fmount: Yeah, makes sense, we can return earlier if the object isn't there `if k8s_errors.IsNotFound(err) {` and we must fail if `err != nil` because we should catch problematic status here.
I'll try to push an update with this logic.

controllers/cinderapi_controller.go
fmount: +1, pushing an update soon with the additional error check we discussed in the previous comment. As per @Akrog comment, I guess it's enough to catch a problematic behavior that should be reported.

controllers/cinderapi_controller.go
```diff
@@ -253,20 +253,19 @@ func (r *CinderAPIReconciler) reconcileDelete(ctx context.Context, instance *cin
...
-				ServiceDescription: ksSvc["desc"],
-				Enabled:            true,
-				APIEndpoints:       instance.Status.APIEndpoints[ksSvc["name"]],
-				ServiceUser:        instance.Spec.ServiceUser,
-				Secret:             instance.Spec.Secret,
-				PasswordSelector:   instance.Spec.PasswordSelectors.Service,
+
+			ks, err := keystone.GetKeystoneServiceWithName(ctx, helper, ksSvc["name"], instance.Namespace)
+
+			if err != nil {
```
fmount, line 259: @abays @Akrog Does this check like this make sense to you?
I guess I can't `if err != nil && k8s_errors.IsNotFound(err) {` because we return `err` or `nil` on the two if statements.
Not an expert here so I'd like to have your input!

abays: The way you now have it is the usual approach we've taken on past operators, so this looks good

abays (approved):
/lgtm

Akrog (approved):
Thanks for the updates, lgtm

PR #18, opened 2022-09-12
Always attempt to remove finalizer on delete

abays:
We need to make sure that the finalizer is removed from the CinderAPI CR regardless of what the current state of Cinder is within Keystone

### DIFF
--- a/controllers/cinderapi_controller.go
+++ b/controllers/cinderapi_controller.go
@@ -256,18 +256,20 @@ func (r *CinderAPIReconciler) reconcileDelete(ctx context.Context, instance *cin
 
 			ks, err := keystone.GetKeystoneServiceWithName(ctx, helper, ksSvc["name"], instance.Namespace)
 
-			if err != nil {
-				if k8s_errors.IsNotFound(err) {
-					return ctrl.Result{}, nil
-				}
+			if err != nil && !k8s_errors.IsNotFound(err) {
 				return ctrl.Result{}, err
 			}
 
-			ksSvcObj := keystone.NewKeystoneService(ks.Spec, instance.Namespace, map[string]string{}, 10)
-			err = ksSvcObj.Delete(ctx, helper)
+			// Could get here if there was an error and it was "not found", so need to check again
+			// (we do it this way because we want to make sure the "RemoveFinalizer" call is
+			// executed even if the KeystoneService no longer exists)
+			if !k8s_errors.IsNotFound(err) {
+				ksSvcObj := keystone.NewKeystoneService(ks.Spec, instance.Namespace, map[string]string{}, 10)
+				err = ksSvcObj.Delete(ctx, helper)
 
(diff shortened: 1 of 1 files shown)


controllers/cinderapi_controller.go
```diff
@@ -256,18 +256,20 @@ func (r *CinderAPIReconciler) reconcileDelete(ctx context.Context, instance *cin
...
+			if err != nil && !k8s_errors.IsNotFound(err) {
 				return ctrl.Result{}, err
 			}
 
-			ksSvcObj := keystone.NewKeystoneService(ks.Spec, instance.Namespace, map[string]string{}, 10)
-			err = ksSvcObj.Delete(ctx, helper)
+			// Could get here if there was an error and it was "not found", so need to check
+			// (we do it this way because we want to make sure the "RemoveFinalizer" call is
+			// executed even if the KeystoneService no longer exists)
+			if err == nil && ks != nil {
```
stuggi, line 266: we'd have to also change https://github.com/openstack-k8s-operators/keystone-operator/blob/master/pkg/external/funcs.go#L247 to return nil for the ks if not found.

abays: Ah yeah, forgot about that.  I'll do that in another PR now.

abays: Actually, for the moment, let's just check here for `ks != nil` only.  The required changes to the Keystone operator have ripple effects that I want to make sure I understand.

controllers/cinderapi_controller.go
```diff
@@ -256,18 +256,20 @@ func (r *CinderAPIReconciler) reconcileDelete(ctx context.Context, instance *cin
...
+			if err != nil && !k8s_errors.IsNotFound(err) {
 				return ctrl.Result{}, err
 			}
 
-			ksSvcObj := keystone.NewKeystoneService(ks.Spec, instance.Namespace, map[string]string{}, 10)
-			err = ksSvcObj.Delete(ctx, helper)
+			// Could get here if there was an error and it was "not found", so need to check again
+			// (we do it this way because we want to make sure the "RemoveFinalizer" call is
+			// executed even if the KeystoneService no longer exists)
+			if !k8s_errors.IsNotFound(err) {
```
abays, line 266: If `err` is `nil`, this returns `false`: https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsNotFound

fmount: Thanks @abays, I tried this patch and it works as expected in a local environment, so +1 on my side.
I'll do the same in [1] and test.

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/34

fmount (approved):

controllers/cinderapi_controller.go
abays: Sounds good!

PR #19, opened 2022-09-12
Fix Ceph templates

Akrog:
When we updated the Ceph backend to use lib-common on commit 00a7370b3c7453406ab24c9e965cce26819a7769 we missed a couple of templates.

This patch updates the missing templates and fixes the Ceph backend.


templates/cinder/config/cinder-volume-config.json
```diff
@@ -21,7 +21,7 @@
     },
     {
       "source": "/var/lib/config-data/merged/ceph.client.openstack.keyring",
-      "dest": "/etc/ceph/ceph.client.{{ .CephUser }}.keyring",
+      "dest": "/etc/ceph/ceph.client.{{ .User }}.keyring",
```
fmount, line 24: +1

templates/cinder/config/cinder.conf
```diff
@@ -86,7 +86,7 @@ lock_path = /var/lib/cinder/tmp
 volume_backend_name=ceph
```
fmount, line 86: Thanks

fmount (approved):

PR #20, opened 2022-09-13
Improve the Cinder Probes

Akrog:
In this PR we aim to improve the Cinder probes doing 2 things:

- Removing readiness probes from non HTTP servers (backup, scheduler,  volume) since they don't need to be removed from any load balancer.

- Replace the liveness command probes from non HTTP servers with a HTTP probe using a custom HTTP server that queries the DB directly.

This is a big improvement for small systems, as it is more efficient.


templates/cinder/bin/kolla_extend_start
```diff
@@ -0,0 +1,7 @@
+#!/bin//bash
```
abays, line 1: Is there an extra `/` here?

abays (commented):
I can't speak to the contents of the Python script, but everything else is +1 from me

templates/cinder/bin/healthcheck.py
```diff
@@ -0,0 +1,155 @@
...
+# For volume services all enabled backends must be up to return 200, so it is
+# recommended to use a different pod for each backend to avoid one backend
+# affecting others.
+
+from http import server
+import sys
+
+from oslo_config import cfg
+
+from cinder import context
```
olliewalsh, line 38: Slight concern that importing openstack python modules in health checks could result in tight coupling of the operator version and the OSP release version. In the mid to long term it may be best to push this upstream

templates/cinder/bin/kolla_extend_start
```diff
@@ -0,0 +1,7 @@
...
+
+if [[ -n "$HEALTHCHECK_SERVICE" ]]; then
+    echo "Cinder Healthcheck Server started"
+    # Dump heartbeat request and error output for now
+    /usr/local/bin/container-scripts/healthcheck.py $HEALTHCHECK_SERVICE &>/dev/null &
```
olliewalsh, line 6: Would a sidecar container be better than a background bash job?

templates/cinder/config/cinder-backup-config.json
```diff
@@ -7,6 +7,12 @@
       "owner": "root:cinder",
       "perm": "0640"
     },
+    {
+      "source": "/usr/local/bin/container-scripts/kolla_extend_start",
+      "dest": "/usr/local/bin/kolla_extend_start",
+      "owner": "root:cinder",
+      "perm": "0740"
```
olliewalsh, line 14: 0740 looks odd. I expect 0640 is all we need if it's sourced by the main kollla script

olliewalsh: https://github.com/openstack/kolla/blob/master/docker/base/start.sh#L20

Akrog: Yes, these are wrong, thanks for catching it.

templates/cinder/bin/kolla_extend_start
Akrog: I'm ok with a sidecar if that's more in line with our design (which probably is).  I just felt that at some point we may want to control the number of pods and containers per pod we require.

Akrog: Would a failure on the sidecar container restart the whole pod or just the sidecar container?

templates/cinder/bin/healthcheck.py
olliewalsh: Maybe upstream in the tripleo images?

templates/cinder/bin/kolla_extend_start
olliewalsh: Just the sidecar container AFAIK

olliewalsh: (I took "failure on the sidecar container" to mean it exits with non-zero)

templates/cinder/bin/healthcheck.py
Akrog: It could be relevant there, don't know enough about their criteria to be sure.

templates/cinder/bin/kolla_extend_start
Akrog: Updated the code to use the sidecar healthcheck container pattern.

pkg/cinderbackup/statefulset.go
```diff
@@ -113,12 +111,26 @@ func StatefulSet(
 							SecurityContext: &corev1.SecurityContext{
 								RunAsUser: &runAsUser,
 							},
-							Env:            env.MergeEnvs([]corev1.EnvVar{}, envVars),
-							VolumeMounts:   GetVolumeMounts(),
-							Resources:      instance.Spec.Resources,
-							ReadinessProbe: readinessProbe,
-							LivenessProbe:  livenessProbe,
-							StartupProbe:   startupProbe,
+							Env:           env.MergeEnvs([]corev1.EnvVar{}, envVars),
```
olliewalsh, line 114: should this use finalEnvVars?

pkg/cinderbackup/statefulset.go
```diff
@@ -113,12 +111,26 @@ func StatefulSet(
...
+							VolumeMounts:  GetVolumeMounts(),
```
olliewalsh, line 115: and volumeMounts

pkg/cindervolume/statefulset.go
```diff
@@ -115,12 +113,26 @@ func StatefulSet(
 								RunAsUser:  &runAsUser,
 								Privileged: &trueVar,
 							},
-							Env:            env.MergeEnvs([]corev1.EnvVar{}, envVars),
-							VolumeMounts:   GetVolumeMounts(),
-							Resources:      instance.Spec.Resources,
-							ReadinessProbe: readinessProbe,
-							LivenessProbe:  livenessProbe,
-							StartupProbe:   startupProbe,
+							Env:           env.MergeEnvs([]corev1.EnvVar{}, envVars),
```
olliewalsh, line 116: finalEnvVars and volumeMounts here too?

templates/cinder/bin/healthcheck.py
```diff
@@ -0,0 +1,162 @@
...
+    CONF(['--config-file', cfg_file], project='cinder')
+
+    HeartbeatServer.initialize_class(binary)
+
+    webServer = server.HTTPServer((HOSTNAME, SERVER_PORT), HeartbeatServer)
+    print(f"Cinder Healthcheck Server started http://{HOSTNAME}:{SERVER_PORT}")
+
+    try:
+        webServer.serve_forever()
+    except KeyboardInterrupt:
```
olliewalsh, line 158: Might be worth adding a SIGTERM handler. IIRC KeyboardInterrupt is only raised for SIGINT. Container stop will send a SIGTERM and (if still running after a timeout) a SIGKILL.

pkg/cinderbackup/statefulset.go
```diff
@@ -113,12 +111,26 @@ func StatefulSet(
...
+								"backup",
+								"/var/lib/config-data/merged/cinder.conf",
+							},
+							Image: instance.Spec.ContainerImage,
+							SecurityContext: &corev1.SecurityContext{
+								RunAsUser: &runAsUser,
+							},
+							Env:          finalEnvVars,
+							VolumeMounts: volumeMounts,
+							Resources:    instance.Spec.Resources,
```
abays, line 133: I'm ambivalent as to whether we should set the probe pod's `Resources` to the value provided in the CR's spec.  Maybe we want to hardcode something here because this probe would be pretty static in its CPU/mem consumption?  Or perhaps leave it out all together and let k8s decide or use a default?

pkg/cinderbackup/statefulset.go
```diff
@@ -62,30 +57,33 @@ func StatefulSet(
 	}
 
 	args := []string{"-c"}
+	// The liveness probe is disabled when debugging the service, but the HTTP
+	// liveness checker server is always started.
```
abays, line 61: We could leave the probe container out when the CR has debug set for the service, but it's not really important.

templates/cinder/bin/healthcheck.py
Akrog: done

Akrog:
Latest update is to resolve merge conflicts with https://github.com/openstack-k8s-operators/cinder-operator/pull/33

pkg/cinderbackup/statefulset.go
```diff
@@ -63,24 +58,35 @@ func StatefulSet(
...
-			Command: []string{
-				"/usr/local/bin/container-scripts/healthcheck.sh",
-				"cinder-backup",
-			},
+		// Use the HTTP probe now that we have a simple server running
+		livenessProbe.HTTPGet = &corev1.HTTPGetAction{
+			Port: intstr.FromInt(8080),
+		}
+		startupProbe.HTTPGet = livenessProbe.HTTPGet
+		// Probe doesn't run kolla_set_configs, so cinder.conf is in its
```
olliewalsh, line 82: Any particular reason to not run kolla_set_config? It might be necessary to set the correct owner/mode if not running as root

pkg/cinderbackup/statefulset.go
```diff
@@ -123,12 +132,21 @@ func StatefulSet(
...
+							Resources:     instance.Spec.Resources,
+							LivenessProbe: livenessProbe,
+							StartupProbe:  startupProbe,
+						},
+						{
+							Name:    "probe",
+							Command: probeCommand,
+							Image:   instance.Spec.ContainerImage,
+							SecurityContext: &corev1.SecurityContext{
+								RunAsUser: &runAsUser,
```
olliewalsh, line 146: Could this run as the cinder user?

pkg/cinderbackup/statefulset.go
Akrog: If we run `kolla_set_config` with a non-root user it will fail due to permissions, as it needs to set some configurations to be owned by `root`.

pkg/cinderbackup/statefulset.go
Akrog: I can run it, but currently cinder user is not 1000, it is 42407, same as its group.
I feared that setting this magic number may break us in the future since the container may change this at any point, but I dig a bit more after your comment and based on [where the magic number comes from](https://github.com/openstack/kolla/blob/0d4e8e6c11947e07953121d6009809db69f4965b/kolla/common/users.py#L40-L42) I don't think we have much to fear so I'll update the patch.

olliewalsh (approved):
/lgtm

PR #21, opened 2022-09-13
Use latest keystone-operator API

Akrog:
pkg/external has been moved to api/v1beta1 so adapt the cinder-operator to this new change.

The keystone change was https://github.com/openstack-k8s-operators/keystone-operator/pull/90


Akrog:
Need this PR to unblock the gate, because we can see how it fails on this earlier PR https://github.com/openstack-k8s-operators/cinder-operator/pull/20

To me it looks like tests are ignoring the pinned version.

PR #33, opened 2022-09-16
Reduce memory consumption in Cinder services

Akrog:
This patch reduces memory usage on the Cinder Volume and Backup services by tuning glibc.

The specific tuning consist on disabling the per thread arenas and disabling dynamic thresholds.

The Cinder Backup service suffers from high water mark memory usage and uses excessive memory.  As an example just after 10 restore operations the service uses almost 1GB of RAM and does not ever free it afterwards. With this patch the memory consumption of the service is reduced down to almost 130MB and memory consumption during runtime is also halved.

This patch is the equivalent to OpenStack patches:

- DevStack: https://review.opendev.org/c/openstack/devstack/+/845805
- TripleO: https://review.opendev.org/c/openstack/tripleo-common/+/845807

Information on glibc tuning is available at:
https://www.gnu.org/software/libc/manual/html_node/Memory-Allocation-Tunables.html

### DIFF
--- a/pkg/cinderbackup/statefulset.go
+++ b/pkg/cinderbackup/statefulset.go
@@ -86,6 +86,14 @@ func StatefulSet(
 	envVars["KOLLA_CONFIG_STRATEGY"] = env.SetValue("COPY_ALWAYS")
 	envVars["CONFIG_HASH"] = env.SetValue(configHash)
 
+	// Tune glibc for reduced memory usage and fragmentation using single malloc arena for all
+	// threads and disabling dynamic thresholds to reduce memory usage when using native threads
+	// directly or via eventlet.tpool
+	// https://www.gnu.org/software/libc/manual/html_node/Memory-Allocation-Tunables.html
+	envVars["MALLOC_ARENA_MAX"] = env.SetValue("1")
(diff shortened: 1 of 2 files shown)


dprince:
patch looks good (lgtm). @Akrog wondering if we need to do similar for other openstack services? Also, should we be setting a pod request/limit alongside of these to further fine tune the statefulsets?

Akrog:
> patch looks good (lgtm). @Akrog wondering if we need to do similar for other openstack services? Also, should we be setting a pod request/limit alongside of these to further fine tune the statefulsets?

Originally I did a good amount of testing for Cinder services and I clearly saw how it fixed the watermark memory issue and how it reduce memory consumption.

[My original devstack patch was changing all devstack services](https://review.opendev.org/c/openstack/devstack/+/845805/3), but we observed that it [didn't seem to help other services](https://review.opendev.org/c/openstack/devstack/+/845805/comments/93a9030a_f5bb284e) so we left it only for Cinder.

We may want to set resource limits at some point, but I'd rather not do it now since it's not that easy to determine. For example for Cinder-Backup it depends on the configured max concurrency and size of the chunks, and for Cinder-Volume I don't really know what are reasonable values.

abays (approved):
/lgtm

PR #36, opened 2022-09-19
Using internal Glance URL

Akrog:
On commit a6aaee5063adb533226d8e2a25b742c42ebdd158 we intended to change the way we get the Glance server addresses from using a hardcoded address to using the Keystone internal Glance URL catalog entry, but due to a mistake the service is using the public one instead (default when value not defined).

This patch fixes this and Cinder will use the intended internal URL instead.

This fixes the mistake from https://github.com/openstack-k8s-operators/cinder-operator/pull/32


stuggi (approved):
/lgtm

PR #37, opened 2022-09-19
Update makefile

Akrog:
These patches change our Makefile to make our life easier as developers:

- Removes the need to have the `kubectl` command, using `oc` instead.
- Add support to pushing to insecure registries
- `make deploy` will work as it is after using the meta-operator.
- `install/uninstall` targets will not only take care of CRDs but also RBACs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,12 +149,12 @@ ifndef ignore-not-found
 endif
 
 .PHONY: install
-install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | kubectl apply -f -
+install: manifests kustomize ## Install CRDs and RBAC into the K8s cluster specified in ~/.kube/config.
+	$(KUSTOMIZE) build config/dev | kubectl apply -f -
 
 .PHONY: uninstall
-uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
-	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
+uninstall: manifests kustomize ## Uninstall CRDs and RBAC from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
+	$(KUSTOMIZE) build config/dev | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
 
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,8 @@ BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
 # BUNDLE_GEN_FLAGS are the flags passed to the operator-sdk generate bundle command
 BUNDLE_GEN_FLAGS ?= -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
 
+VERIFY_TLS ?= true
+
 # USE_IMAGE_DIGESTS defines if images are resolved via tags or digests
 # You can enable this value if you would like to use SHA Based Digests
 # To enable set flag to true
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -47,7 +49,7 @@ ifeq ($(USE_IMAGE_DIGESTS), true)
 endif
 
 # Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+IMG ?= quay.io/openstack-k8s-operators/cinder-operator:latest
```
dprince, line 52: So this is actually flagged in some of our build automation. https://github.com/openstack-k8s-operators/cinder-operator/blob/master/.github/create_bundle.sh#L50 The same default exists in other operators

dprince:
@Akrog thanks for this. Looking at some of the contents here and wondering if we do it for Cinder would we apply similar elsewhere

Makefile
```diff
@@ -131,21 +133,21 @@ ifndef ignore-not-found
...
 .PHONY: deploy
 deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | kubectl apply -f -
+	$(KUSTOMIZE) build config/default | oc apply -f -
 
 .PHONY: undeploy
 undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
-	$(KUSTOMIZE) build config/default | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
+	$(KUSTOMIZE) build config/default | oc delete --ignore-not-found=$(ignore-not-found) -f -
```
dprince, line 150: As we periodically sync a majority of these Makefiles with operator-sdk it is somewhat useful to minimize changes. Not saying we can't default to 'oc' but many of us soft link kubectl -> oc anyways to make stuff like this just work (they are mostly the same)

Akrog: Somebody was telling this yesterday in Slack, so I think instead of renaming it the `Makefile` should make a symlink if kubectl doesn't exist to avoid, like you say, having it changed by the operator-sdk.

I'll look into it.

Akrog:
> @Akrog thanks for this. Looking at some of the contents here and wondering if we do it for Cinder would we apply similar elsewhere

I admit I haven't looked at other projects, I started making these changes when I worked on a [_Getting started guide_](https://github.com/Akrog/cinder-operator/blob/5dff04057cdae90660f11f71798e24731348de12/README.md) for the project and I noticed some inconsistencies.

Makefile
Akrog: I changed the first patch so now it creates a symlink if `kubectl` doesn't exist instead of removing the `kubectl` references. 
I think this aligns better with what you are suggesting.

Makefile
```diff
@@ -38,6 +38,8 @@ BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
 # BUNDLE_GEN_FLAGS are the flags passed to the operator-sdk generate bundle command
 BUNDLE_GEN_FLAGS ?= -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
 
+USE_TLS ?= true
```
olliewalsh, line 41: VERIFY_TLS would be better I think

Makefile
olliewalsh: Is there a danger of a local build pushing this to quay if the user happens to have write permission?

Makefile
Akrog: Yeah, sounds better.

> There are only two hard things in Computer Science: cache invalidation and naming things.
> 
> -- Phil Karlton

Makefile
Akrog: There is, and I had a previos version of the patch where it checked in  the `docker-push` target to prevent it from working if it was trying to push there, but I thought that we could be using it on the GH actions to actually push there...

olliewalsh: GH action already does - https://github.com/openstack-k8s-operators/cinder-operator/actions/runs/3112473499/jobs/5045972880. On the main repo the action will push to quay.io/openstack-k8s-operators, on a fork it will push to quay.io/<quay_username>

Akrog: Oh, but it doesn't use `make docker-push`, so I can add checks in our `Makefile` to prevent it!
Thanks for the reference.  :-)

Akrog: Done

olliewalsh (approved):

PR #38, opened 2022-09-20
Debug improvements

Akrog:
This PR separates `debug` in 2 different structures, one for global things and another for the services themselves.

That way we don't have things like `dbSync` debugging in the services, or `service` for the global.

It also adds the possibility of debugging init containers, which is helpful for newcomers to explore the full process as well as to developers when they want to debug these containers.


abays:
I think this is a helpful change, but I would want to see us try to maintain the same debug pattern across all the operators.  It seems like this would fit elsewhere too, but perhaps we should get others to chime-in with their thoughts as well.

dprince (approved):

PR #39, opened 2022-09-21
Max 1 replica for volume

Akrog:
We currently only support Active-Passive deployments for the cinder-volume service since we haven't resolved yet the whole DLM situation or how to know if the driver supports Active-Active.

For this reason we'll limit the number of replicas supported for the volume backends to 1 through the validations.


abays (approved):
/lgtm

ASBishop:
lgtm

PR #43, opened 2022-09-23
WIP: cleaner kolla config

olliewalsh:


templates/cinderbackup/config/kolla-config.json
```diff
@@ -1,5 +1,9 @@
 {
+  {{if .ServiceDebug}}
```
olliewalsh, line 2: Maybe instead of a conditional just have a "ServiceCommand" var that if set will override the default?

pkg/cinder/const.go
```diff
@@ -42,4 +42,11 @@ const (
 
 	// KollaConfigDbSync -
 	KollaConfigDbSync = "/var/lib/config-data/merged/db-sync-config.json"
+
+	// KollaDefaultCommand -
+	KollaDefaultCommand = "/usr/bin/cinder-backup --config-file /usr/share/cinder/cinder-dist.conf --config-file /etc/cinder/cinder.conf"
+	// KollaDefaultProbeCommand -
+	KollaDefaultProbeCommand = "/usr/local/bin/container-scripts/healthcheck.py backup"
```
olliewalsh, line 49: for the api services the runs as root maybe run `sudo -u cinder -E ...`?

olliewalsh: or the healthcheck script could setuid to cinder

Akrog (changes requested):
I'm in agreement with the direction of the implementation, but there is an issue with the actual implementation, as it breaks Ceph deployments.

pkg/cinderbackup/statefulset.go
```diff
@@ -181,7 +147,7 @@ func StatefulSet(
 		OSPSecret:            instance.Spec.Secret,
 		DBPasswordSelector:   instance.Spec.PasswordSelectors.Database,
 		UserPasswordSelector: instance.Spec.PasswordSelectors.Service,
-		VolumeMounts:         GetInitVolumeMounts(),
+		VolumeMounts:         GetInitVolumeMounts(instance.Name),
```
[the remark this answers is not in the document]
olliewalsh: Can stay with the previous name. The mount name was `config-data-custom` but will now provide more than the custom .conf so just wanted to make that a bit clearer.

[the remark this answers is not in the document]
Akrog: If you refer to how it now provides the command to run as well, then I agree that changing the name from `config-data-custom` to something like `config-data` is better. I'm just not sure of the benefit of prefixing it with the instance name as well.
In any case this is not a blocker, just an observation.

Akrog:
We no longer support the debug struct that required the different commands (we now use `oc debug`), so this PR seems to be no longer needed.

PR #44, opened 2022-09-23
Drop registry.redhat.io login from github action

olliewalsh:
Only required if we are converting image tags to shas and reference registry.redhat.io images.


abays (approved):
/lgtm

PR #47, opened 2022-09-28
Simplify resource cleanup

stuggi:
To cleanup the created resources it is ok to just remove the finalizer which gets set during create. When the owner gets deleted and the owning resource removes the finalizer. k8s does does the cleanup of those resources where the owner is missing and the relevant CR controller calls the cleanup actions, if any.

### DIFF
--- a/controllers/cinderapi_controller.go
+++ b/controllers/cinderapi_controller.go
@@ -257,22 +257,32 @@ func (r *CinderAPIReconciler) reconcileDelete(ctx context.Context, instance *cin
 	if instance.Status.APIEndpoints != nil {
 		for _, ksSvc := range keystoneServices {
 
-			ks, err := keystonev1.GetKeystoneServiceWithName(ctx, helper, ksSvc["name"], instance.Namespace)
-
+			// Remove the finalizer from our KeystoneEndpoint CR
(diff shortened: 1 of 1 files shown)


abays (commented):
Looks like the right thing to me, but I'll wait to approve until @Akrog tests so that the bot doesn't auto-merge

Akrog (approved):
Code looks good and works as intended

controllers/cinderapi_controller.go
```diff
@@ -257,22 +257,32 @@ func (r *CinderAPIReconciler) reconcileDelete(ctx context.Context, instance *cin
...
 				return ctrl.Result{}, err
 			}
 
-			// Could get here if there was an error and it was "not found", so need to check again
-			// (we do it this way because we want to make sure the "RemoveFinalizer" call is
-			// executed even if the KeystoneService no longer exists)
-			if !k8s_errors.IsNotFound(err) {
-				ksSvcObj := keystonev1.NewKeystoneService(ks.Spec, instance.Namespace, map[string]string{}, 10)
-				err = ksSvcObj.Delete(ctx, helper)
+			if err == nil {
```
fmount, line 266: I see we need to explicitly check for `err` here as the previous condition can't be always met because of the `&&`.
(Same thing applies to the service)

fmount (approved):

PR #48, opened 2022-09-28
Fix bundle automation

Akrog:
On commit 44190808a46574c0bdef27d80738713766f627cd I broke the bundle automation when trying to make the `deploy` rule work off the bat not considering (even with a review comment) that there are automation [1] code checking for the specific `controller:latest` string.

In this patch we revert the change made to the `Makefile` file by that patch and create a new rule called `deploy-latest` instead to stay out of the automation way while still providing an easy way for developers to deploy the latest official image without having to enter its location.

The whole commit was not reverted because the namespace part was still wanted.

[1]: https://github.com/openstack-k8s-operators/cinder-operator/blob/master/.github/create_bundle.sh#L50

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,8 +49,8 @@ ifeq ($(USE_IMAGE_DIGESTS), true)
 endif
 
 # Image URL to use all building/pushing image targets
+IMG ?= controller:latest
 DEFAULT_IMG ?= quay.io/openstack-k8s-operators/cinder-operator:latest
-IMG ?= $(DEFAULT_IMG)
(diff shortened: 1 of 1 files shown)


Akrog:
Abandoning this as it looks like the bundle automation was not actually broken and [the mentioned code](https://github.com/openstack-k8s-operators/cinder-operator/blob/master/.github/create_bundle.sh#L50) [can be removed](https://github.com/openstack-k8s-operators/cinder-operator/pull/50)

PR #50, opened 2022-09-28
Drop controller:latest logic from create_bundle.sh

dprince:
This would no longer get triggered as our default Makefile image is now set to a fully qualified URL (quay.io....)


Akrog:
Does look good to me, but I'm not familiar enough with the automation to merge it.

Akrog (approved):

PR #51, opened 2022-09-29
Share os-brick locks between OpenStack services

Akrog:
os-brick file locks need to be shared across all the OpenStack services
that use os-brick and are running on the same host, if we don't there
will be race conditions that make attach/detach operation unexpectedly
fail.

Services that need to share this locks location are cinder-volume,
cinder-backup, and glance when using cinder as a backend.  Nova compute
would also need to share them in an HCI environment, but that won't be
the case here because nova-compute won't be running in OpenShift.

This patch uses the host's `/var/locks` directory to share these locks.

In OSP17 there is a single lock path configuration option for cinder and
os-brick locks, but in OSP18 there is an os-brick specific configuration
option that can be used.

To make the operator compatible with OSP17 and OSP18 cinder-volume and
cinder-backup pods will "share" both file lock locations
`/var/locks/openstack/os-brick` and `/var/locks/openstack/cinder`
through the host.

### DIFF
--- a/pkg/cinder/dbsync.go
+++ b/pkg/cinder/dbsync.go
@@ -55,16 +55,16 @@ func DbSyncJob(instance *cinderv1beta1.Cinder, labels map[string]string) *batchv
 								RunAsUser: &runAsUser,
 							},
 							Env:          env.MergeEnvs([]corev1.EnvVar{}, envVars),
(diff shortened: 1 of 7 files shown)


abays (approved):
/lgtm

PR #56, opened 2022-10-12
Show overall Cinder status in CLI

abays:
`oc get cinder` will now surface the status (following the pattern used by other operators)

### DIFF
--- a/api/bases/cinder.openstack.org_cinders.yaml
+++ b/api/bases/cinder.openstack.org_cinders.yaml
@@ -15,7 +15,16 @@ spec:
     singular: cinder
   scope: Namespaced
   versions:
-  - name: v1beta1
+  - additionalPrinterColumns:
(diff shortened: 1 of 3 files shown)


api/bases/cinder.openstack.org_cinders.yaml
```diff
@@ -15,7 +15,16 @@ spec:
     singular: cinder
   scope: Namespaced
   versions:
-  - name: v1beta1
+  - additionalPrinterColumns:
+    - description: Status
+      jsonPath: .status.conditions[0].status
```
fmount, line 20: I see this is consistent with the other operators (e.g. glance/keystone): as long as it's the pattern used by all the operators, lgtm.

fmount (approved):

Akrog:
Thanks for this!  :-)

PR #65, opened 2022-11-06
Extra volumes support to cinder-operator

fmount:
This patch aligns the code with the lib-common ExtraVolumes change, where the ExtraVolumes support is provided for each storage component. Additional extra mounts can be defined in the top level cinder CRD and passed to the GetVolume functions provided by lib-common. Then, according to the Propagation policy, the extraMount will be (or not be) propagated to the service deployed by the cinder-operator.

Depends-on: openstack-k8s-operators/lib-common#88
Depends-on: openstack-operator#38


templates/cinder/config/cinder-volume-config.json
```diff
@@ -12,18 +12,6 @@
       "dest": "/etc/cinder/nfs_shares",
       "owner": "root:cinder",
       "perm": "0640"
-    },
-    {
```
fmount, line 16: @Akrog I see a problem to keep the old ceph mechanism coexisting with extra voumes: in particular if I propagate an extra volume in `/etc/ceph`, via somethin like the following:

```
 extraMounts:
      extraVol:
        - propagation:
          - CinderVolume
          extraVolType: Ceph
          volumes:
          - name: ceph
            projected:
              sources:
              - secret:
                  name: ceph-client-conf
          mounts:
          - name: ceph
            mountPath: "/etc/ceph"
            readOnly: false
```

It won't take effect because when the pod starts (cinder-volume in this specific case), `kolla` overwrite/replace the content of the mounted path (even "merge": true doesn't solve the problem) [1].
Not sure you have a suggestion on how to skip the files copy under certain conditions.

I'm thinking about just removing the old ceph way, and rely entirely on this feature once it merges.
 
 [1] https://docs.openstack.org/kolla/ussuri/admin/kolla_api.html

Akrog: I agree 100%. We should remove the old mechanism.
Fortunately we don't have to worry about backward compatibility at this stage and can do breaking changes  :-)

fmount:
@Akrog as long as we import [1] and [2] from `meta-operator` we should be ready to go with both `Cinder` and `Glance`.
Only after this merges we can update  [3] which requires the new cinder operator version in `go.mod` that will allow the CI to pass (and resolve `cinder.Spec.ExtraMounts` in this particular case).
Do we want to wait more (e.g., a recording with the current status) before merging or are you ok with this plan?


[1] https://github.com/fmount/cinder-operator/blob/master/controllers/cinder_controller.go#L48
[2] https://github.com/fmount/cinder-operator/blob/master/main.go#L41
[3] https://github.com/openstack-k8s-operators/openstack-operator/pull/38

api/v1beta1/cinderapi_types.go
```diff
@@ -137,3 +141,4 @@ func init() {
 func (instance CinderAPI) IsReady() bool {
 	return instance.Status.ReadyCount >= 1
 }
+
```
Akrog, line 144: Unnecessary new line at the end of the file

pkg/cinder/dbsync.go
```diff
@@ -55,16 +57,16 @@ func DbSyncJob(instance *cinderv1beta1.Cinder, labels map[string]string) *batchv
 								RunAsUser: &runAsUser,
 							},
 							Env:          env.MergeEnvs([]corev1.EnvVar{}, envVars),
-							VolumeMounts: GetVolumeMounts(false),
+							VolumeMounts: GetVolumeMounts(false, []cinderv1beta1.CinderExtraVolMounts{}, []storage.PropagationType{storage.DBSync}),
```
Akrog, line 60: I think the `[]storage.PropagationType{storage.DBSync})` should be a constant that we only define once, right?

And maybe `[]cinderv1beta1.CinderExtraVolMounts{}` should be a variable in this method that we can use in all 3 instances where it is used.

pkg/cinderapi/volumes.go
```diff
@@ -23,22 +25,22 @@ func GetVolumes(parentName string, name string) []corev1.Volume {
 		},
 	}
 
-	return append(cinder.GetVolumes(parentName, false), backupVolumes...)
+	return append(cinder.GetVolumes(parentName, false, extraVol, []storage.PropagationType{cinder.Cinder, cinder.CinderAPI}), backupVolumes...)
```
Akrog, line 28: Shouldn't `[]storage.PropagationType{cinder.Cinder, cinder.CinderAPI}` be a constant we reuse all over this file?

pkg/cinderbackup/volumes.go
```diff
@@ -23,22 +25,22 @@ func GetVolumes(parentName string, name string) []corev1.Volume {
 		},
 	}
 
-	return append(cinder.GetVolumes(parentName, true), backupVolumes...)
+	return append(cinder.GetVolumes(parentName, true, extraVol, []storage.PropagationType{cinder.Cinder, cinder.CinderBackup}), backupVolumes...)
```
Akrog, line 28: Same comment about making `storage.PropagationType` a constant.

pkg/cinderscheduler/volumes.go
```diff
@@ -23,22 +25,22 @@ func GetVolumes(parentName string, name string) []corev1.Volume {
 		},
 	}
 
-	return append(cinder.GetVolumes(parentName, false), schedulerVolumes...)
+	return append(cinder.GetVolumes(parentName, false, extraVol, []storage.PropagationType{cinder.Cinder, cinder.CinderScheduler}), schedulerVolumes...)
```
Akrog, line 28: Same comment about making storage.PropagationType a constant.

pkg/cindervolume/volumes.go
```diff
@@ -33,29 +37,35 @@ func GetVolumes(parentName string, name string) []corev1.Volume {
 		},
 	}
 
-	return append(cinder.GetVolumes(parentName, true), volumeVolumes...)
+	// Set the propagation levels for CinderVolume, including the backend name
+	propagation := []storage.PropagationType{cinder.Cinder, cinder.CinderVolume, storage.PropagationType(strings.TrimPrefix(name, "cinder-volume-"))}
```
Akrog, line 41: Same comment about making storage.PropagationType a constant.

Akrog (changes requested):

api/v1beta1/cinderapi_types.go
fmount: Done

pkg/cinder/dbsync.go
fmount: Right, they're basically constants or variables that can be reused all over the code, hence I moved the first to the `const` file, while the second it's just a regular variable that can be reused many times in the same function.

pkg/cinderapi/volumes.go
fmount: Done

pkg/cinderscheduler/volumes.go
fmount: Done

pkg/cindervolume/volumes.go
fmount: Done, we have a basic propagation over `Cinder` (global) and `CinderVolume` (group), and we can `append` the instance to the variable because it's dynamically generated.

pkg/cinderbackup/volumes.go
fmount: Done

dprince (approved):

pkg/cinder/dbsync.go
```diff
@@ -55,16 +57,16 @@ func DbSyncJob(instance *cinderv1beta1.Cinder, labels map[string]string) *batchv
...
 					},
-					Volumes: GetVolumes(instance.Name, false),
+					Volumes: GetVolumes(instance.Name, false, dbSyncExtraMounts, DbsyncPropagation),
 				},
 			},
 		},
 	}
 
-	job.Spec.Template.Spec.Volumes = GetVolumes(ServiceName, false)
+	job.Spec.Template.Spec.Volumes = GetVolumes(ServiceName, false, dbSyncExtraMounts, DbsyncPropagation)
```
abays, line 69: I just realized, I think, that this line is unnecessary given that we already have it in L63 (or vice-versa).  63 uses `instance.Name` and this line uses `ServiceName` as the first arg, though.  Not sure off the top of my head which one is the right choice.

PR #78, opened 2022-11-24
Logging: Use ISO8601 timestamps

Akrog:
This patch disables the zap development logging mode that was the
default in the Cinder operator.

The new code does not set any zap options and only expects to read
options from the command line.

For development purposes it is convenient to run logging not only in
development mode, but also increase the log levels and change the time
encoding from epoc to iso8601 which is easier to read for humans, so 2
new make targets are added, `run-dev` and `docker-build-dev` which are
similar to the `run` and `docker-build` targets.

The `run-dev` target is different from the standard `run` because it
will only build the operator if the binary doesn't exist.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,6 @@ CI_TOOLS_REPO
 # generated workspace file
 go.work
 go.work.sum
+
+# Hack files that get modified
+hack/Dockerfile.devel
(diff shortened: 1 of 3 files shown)


Akrog:
This has been fixed somewhere else.

PR #77, opened 2022-11-22
Docs

Akrog:
This PR includes a series of patches to improve the cinder operator documentation to include:

- Getting started guide
- Contributor guidelines
- Using external dependencies
- Debugging guide

PR #72, opened 2022-11-14
Lint improvements

gibizer:
1. ensure that normal linting runs on the `api` module as well
2. introduce operator-lint [1] make target and run it in github workflow to check for operator development specific rule violations.
3. fix such violations

**Please note that this change makes Secret and DatabaseInstance Required while it was Optional before.**

[1] https://github.com/gibizer/operator-lint

PR #79, opened 2022-11-25
Add go.work to deal with multi module dependencies

fmount:
This patch aligns the cinder-operator to the work already done in both glance and openstack operators where go.work is added to deal with multi module dependencies and run linters on API module.


fmount:
@Akrog with this change to the `Makefile` I'm able to just add:

```
 go mod edit -replace github.com/openstack-k8s-operators/lib-common/modules/storage=../lib-common/modules/storage
 ```

and run `make` without failing w/ `invalid type` in the auto-generated zz code.

File: Makefile [hunk 1]
```diff
@@ -272,20 +273,31 @@ get-ci-tools:
...
 golint: get-ci-tools
-	PATH=$(GOBIN):$(PATH); $(CI_TOOLS_REPO_DIR)/test-runner/golint.sh
+	GOWORK=off PATH=$(GOBIN):$(PATH); $(CI_TOOLS_REPO_DIR)/test-runner/golint.sh
+	GOWORK=off PATH=$(GOBIN):$(PATH); $(CI_TOOLS_REPO_DIR)/test-runner/golint.sh ./api
+
+.PHONY: gowork
+gowork: ## Generate go.work file to support our multi module repository
+	test -f go.work || go work init
+	go work use .
+	go work use ./api
```
Akrog, line 303: -1: I would add the go.work directly in the repo instead of having this line

Makefile
```diff
@@ -272,20 +273,31 @@ get-ci-tools:
 
 # Run go fmt against code
 gofmt: get-ci-tools
-	$(CI_TOOLS_REPO_DIR)/test-runner/gofmt.sh
+	GOWORK=off $(CI_TOOLS_REPO_DIR)/test-runner/gofmt.sh
```
Akrog, line 276: -1: This should be overwritable with an env variable

Akrog (changes requested):
Instead of `go mod edit -replace` like we suggest I think it's better to do `go work edit -replace` because that way our local changes will be ignored by git since go.work is in .gitignore
That's what I'm going to suggest in the docs.

fmount:
> Instead of `go mod edit -replace` like we suggest I think it's better to do `go work edit -replace` because that way our local changes will be ignored by git since go.work is in .gitignore That's what I'm going to suggest in the docs.

Ah nice, I didn't know we couldn't have a `go work edit -replace ..`
That approach works for me, so I guess we can go that way given you're suggesting it in the docs/ PR

Makefile
fmount: So you're suggesting running those 3 commands and put the resulting files (`go.work` and `go.work.sum`) within the repository: am I understanding correctly?
By doing that we can avoid this target here.

Makefile
fmount: ack, makes sense I guess, better an env variable defaulting to `off`, in case you need to change this value in the future (in that case you can change just one line instead of playing with all the occurrences).

Makefile
stuggi: FYI, we added this target to other operator Makefile as the design workspace doc says https://go.googlesource.com/proposal/+/master/design/45713-workspace.md#rationale-the-file

~~~
These go.work files should not be checked into the repositories so that they don‘t override the workspaces users explicitly define. Checking in go.work files could also lead to CI/CD systems not testing the actual set of version requirements on a module and that version requirements among the repository’s modules are properly incremented to use changes in the modules.
~~~

Why we could do different, this was the initial reason why we did not.

gibizer: > Instead of `go mod edit -replace` like we suggest I think it's better to do `go work edit -replace` because that way our local changes will be ignored by git since go.work is in .gitignore That's what I'm going to suggest in the docs.

When a commit depends on another commit in another repository I tend to actually check in and publish the `replace` in go.mod so that when people review the PR they can pull it and run it locally with the correct deps. Sure this means once the dependency merges I need to respin my PR but overall but this is my low tech version of zuul's `Depends-On:`

gibizer:
FYI, have this proposed as part of #72  https://github.com/openstack-k8s-operators/cinder-operator/pull/72/commits/fd557c95779a3bcadd75490c88532f3daf45d59d

Makefile
fmount: +1 and thanks @gibizer for sharing. The whole point of having `go.work` (in my recent experience) is to make the `go mod edit -replace` working properly when a local `lib-common` is referenced in the project.
I didn't find a nice way to make it working without creating a go workspace.
It would be great (in the next future) having our CI able to checkout the dependencies and run a job which takes into account the PR dependencies.

fmount:
> FYI, I have proposed a very similar commit as part of #72 [fd557c9](https://github.com/openstack-k8s-operators/cinder-operator/commit/fd557c95779a3bcadd75490c88532f3daf45d59d)

Sorry @gibizer, I didn't realize you already had a PR for the same thing. I'm ok to close this one and move the conversation on your change

gibizer:
> > FYI, I have proposed a very similar commit as part of #72 [fd557c9](https://github.com/openstack-k8s-operators/cinder-operator/commit/fd557c95779a3bcadd75490c88532f3daf45d59d)
> 
> Sorry @gibizer, I didn't realize you already had a PR for the same thing. I'm ok to close this one and move the conversation on your change

No worries. This can be merged separately and I can drop the same thing from my series.

Makefile
Akrog: @stuggi Iiuc that paragraph is referring to not checking in the `go.work` with actual replacements, which is also what I'm suggesting.

By having the `go.work` with only the `.` and `./api` in the repo and excluding it with `.gitignore` we avoid anyone from inadvertently submitting it.

@gibizer I myself don't like the PRs having the `go.mod` changes included.  I have my own python script where I can tell it the PR and it does the replacing automatically. For example for the extra volume stuff where we use the lib-common PR#88 and openstack-operator PR#38 I just run:

```
hack/replace.py lib-common=88 openstack-operator=38
```

And that makes the right replacements in `go.work` for me.

It should be pretty straightforward building the `Depend-On` with that, because the script also accepts local directories and even specific branches of repos, where repos can be long or short version.  For example the short version for the previous one would be:

```
hack/replace.py lib-common=fmount/lib-common@extra_volumes openstack-operator=fmount/openstack-operator@extra_volumes
```

Makefile
Akrog: I was referring to an external env variable that changes the default to `off`:

```
GOWORK ?= off
```

Makefile
gibizer: > @gibizer I myself don't like the PRs having the `go.mod` changes included. I have my own python script where I can tell it the PR and it does the replacing automatically. For example for the extra volume stuff where we use the lib-common PR#88 and openstack-operator PR#38 I just run:
> 
> ```
> hack/replace.py lib-common=88 openstack-operator=38
> ```
> 
> And that makes the right replacements in `go.work` for me.
> 
> It should be pretty straightforward building the `Depend-On` with that, because the script also accepts local directories and even specific branches of repos, where repos can be long or short version. For example the short version for the previous one would be:
> 
> ```
> hack/replace.py lib-common=fmount/lib-common@extra_volumes openstack-operator=fmount/openstack-operator@extra_volumes
> ```

I guess it is matter of taste. I like my way as it makes explicit for the reader of the PR that there is a dependency to a specified version and location (i.e.xyz branch of the fork).
By all means I'm supportive to automate dependency handling via some scripting.

Akrog: I like to have it explicitly stated in the commit message (like with th e`Depends-on` you mention).

stuggi: > @stuggi Iiuc that paragraph is referring to not checking in the `go.work` with actual replacements, which is also what I'm suggesting.
> 
> By having the `go.work` with only the `.` and `./api` in the repo and excluding it with `.gitignore` we avoid anyone from inadvertently submitting it.
> 

right, we definitely want to have it in `.gitignore` to prevent people to add it to commits. Having a default checked in is something we could do, but its not common practice. So far I have not found other projects which do this. In our case for the main and api sub module its probably ok to add the go.work to the repo as they depend on each other.

Makefile
fmount: yes, same thing I had in mind.

Makefile
fmount: Ok, I see @stuggi's point of view and in general I agree about not having `go.work` and `go.work.sum` in the repos.
However, I understand that PR #77  mentioned by @Akrog introduces a series of scripts with the goal of simplify the developer life, automating the way we checkout PRs [1] and set dependencies to `GOWORK` [2] in cinder-operator.
In this case I suspect it's good to maintain consistency with the direction taken by this operator, and we can actually commit a `go.work` file (which is still in `.gitignore`), removing the need to execute the `gowork` target each time.
We might want to keep the `gowork` target in the `Makefile` (it can be explicitly called if needed), but I can remove its execution in the `vet` target (and wherever it occurs).

I'm looking for quorum to move forward with this, so I'd like to have your feedback @stuggi @Akrog 


[1] https://github.com/openstack-k8s-operators/cinder-operator/pull/77/files#diff-adc3e455779ae5d167129a8b2a045ac46868c35ea47fcfb127327f0c3ba7cf45
[2] https://github.com/openstack-k8s-operators/cinder-operator/pull/77/files#diff-58fb17599cdc1f9bd44fbdb460b74f3de822da6a4f5007f0ac97831ee0e4ec68

stuggi: I am fine with the proposal, adding the default `go.work`  with having it in `.gitignore` and remove calling it in the targets.

fmount:
The patch [1] should fix the failure we've seen in CI as a side effect of having a versioned `go.work`.

[1]  https://github.com/openstack-k8s-operators/openstack-k8s-operators-ci/pull/56

gibizer:
> The patch [1] should fix the failure we've seen in CI as a side effect of having a versioned `go.work`.
> 
> [1] [openstack-k8s-operators/openstack-k8s-operators-ci#56](https://github.com/openstack-k8s-operators/openstack-k8s-operators-ci/pull/56)

you can add `GOWORK=off` to the CI targets in the Makefile here as an alternative.

Makefile
```diff
@@ -272,20 +276,31 @@ get-ci-tools:
...
 # Run go fmt against code
 gofmt: get-ci-tools
-	$(CI_TOOLS_REPO_DIR)/test-runner/gofmt.sh
+	GOWORK=$(GOWORK) $(CI_TOOLS_REPO_DIR)/test-runner/gofmt.sh
+	GOWORK=$(GOWORK) $(CI_TOOLS_REPO_DIR)/test-runner/gofmt.sh ./api
 
 # Run go vet against code
 govet: get-ci-tools
-	$(CI_TOOLS_REPO_DIR)/test-runner/govet.sh
+	GOWORK=$(GOWORK) $(CI_TOOLS_REPO_DIR)/test-runner/govet.sh
```
fmount, line 284: @gibizer @stuggi  I already set this here but somehow is failing [1].
I assumed this is related to the fact we added `go.work` within the repo, so I explicitly added the `GOWORK` var within the script [2]. 
By doing this we can remove `GOWORK` here (it can be applied to all the other operators), not sure if you have preferences or this approach has some drawbacks I'm overlooking.

[1] https://github.com/openstack-k8s-operators/cinder-operator/actions/runs/3573061209/jobs/6006679452
[2] https://github.com/openstack-k8s-operators/openstack-k8s-operators-ci/pull/56/files

gibizer: Ahh I always forget that our workflow are not calling make targets (they should I think). So you have to add it to the workflows too https://github.com/openstack-k8s-operators/cinder-operator/blob/6240496cc613a1f7843c0960158210d2e9203631/.github/workflows/golangci-lint.yaml#L21-L26

fmount: ... or, according to the review [1], `GOWORK=off` is automatically pushed in the script, preserving the existing behavior, and I just have to remove the  `GOWORK` references in the `Makefile` here, am I right?
@gibizer added you to [1] 

[1] https://github.com/openstack-k8s-operators/openstack-k8s-operators-ci/pull/56

gibizer: Yes, that would work too.

Makefile
```diff
@@ -40,6 +40,9 @@ BUNDLE_GEN_FLAGS ?= -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
 
 VERIFY_TLS ?= true
 
+# GOWORK
+GOWORK ?= off
+
```
Akrog, line 45: Why don't we use the Make `export` construct so that it is present in all the executions?

That would save us from adding GOWORK everywhere in the `Makefile`

```
export GOWORK := $GOWORK
```

Akrog (changes requested):

Makefile
Akrog: If we really don't want to have the go.work file included to be the same as other projects I could modify my scripts to create it when necessary, though we would still want to default `GOWORK` to off

Akrog (changes requested):
We also need to modify Dockerfile to support `GOWORK`

Makefile
```diff
@@ -118,10 +121,11 @@ fmt: ## Run go fmt against code.
 .PHONY: vet
 vet: ## Run go vet against code.
 	go vet ./...
+	go vet ./... ./api/...
```
Akrog, line 124: Doesn't this replace L123?

fmount: This is an interesting point. I assumed that something like [1] is the way to properly run `go vet` both in the main project and the related `./api` tree.

However, running `go list` like the following:

```
> go list ./... ./api/...
github.com/openstack-k8s-operators/cinder-operator
github.com/openstack-k8s-operators/cinder-operator/controllers
github.com/openstack-k8s-operators/cinder-operator/pkg/cinder
github.com/openstack-k8s-operators/cinder-operator/pkg/cinderapi
github.com/openstack-k8s-operators/cinder-operator/pkg/cinderbackup
github.com/openstack-k8s-operators/cinder-operator/pkg/cinderscheduler
github.com/openstack-k8s-operators/cinder-operator/pkg/cindervolume
github.com/openstack-k8s-operators/cinder-operator/api/v1beta1
```
we can see we're able to "discover" both the main tree of the project, but also the `./api/` part with a single command.
This means that you're actually right, L123 is enough to run `go vet` against the entire project. 
This is also proved by the fact that introducing (by purpose) an error within `cinderapi_types.go` I get:
```
> go vet ./... ./api/...
# github.com/openstack-k8s-operators/cinder-operator/api/v1beta1
api/v1beta1/cinderapi_types.go:132:56: syntax error: unexpected aaa at end of statement
```

So yes, I'll change this part just having L123, thanks @Akrog !

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/master/Makefile#L119-L120

fmount:
> We also need to modify Dockerfile to support `GOWORK`

Yes, we might need something like:

```
ARG GOWORK=off
...
...
...
...... CGO_ENABLED=0  GO111MODULE=on GOWORK=${GOWORK} go build ${GO_BUILD_EXTRA_ARGS} -a -o .....
```
within the Dockerfile, and we might need to update the other operators as well.
am I understanding you properly here?

Makefile
fmount: > Why don't we use the Make `export` construct so that it is present in all the executions?
> 
> That would save us from adding GOWORK everywhere in the `Makefile`
> 
> ```
> export GOWORK := $(GOWORK)
> ```

Not sure I'm properly following you here, but as per [1] (if it lands) I can remove all the `GOWORK=$(GOWORK)` occurrences in the `Makefile` and just have `GOWORK ?= off` at the top of the file as you previously suggested.

Then, if you run something like:
```
GOWORK=on make govet
```

you end up seeing something like [2], but if you run:

```
GOWORK=off make govet 
``` 
(or just `make govet`) you'll see something like [3].

IMO this would simplify a lot the `Makefile` and we don't have to deal w/ that variable unless we would like to set it to `on`.

@Akrog does this make sense?


[2] https://paste.opendev.org/show/bLGg1Dx7D4ueNmqIT5kQ/
[3] https://paste.opendev.org/show/b48LCNwZFoOk3DdPKraI/

Akrog: If you just add:
```
GOWORK ?= off
export GOWORK := $(GOWORK)
```
Then you don't need to add `GOWORK=$(GOWORK)` anywhere in this file.
And then we you want to use the go workspace you can do:
```
GOWORK= make build
```
And it will use the current go workspace.

Akrog:
> > We also need to modify Dockerfile to support `GOWORK`
> 
> Yes, we might need something like:
> 
> ```
> ARG GOWORK=off
> ...
> ...
> ...
> ...... CGO_ENABLED=0  GO111MODULE=on GOWORK=${GOWORK} go build ${GO_BUILD_EXTRA_ARGS} -a -o .....
> ```
> 
> within the Dockerfile, and we might need to update the other operators as well. am I understanding you properly here?

Yes, that or adding after `FROM $GOLANG_BUILDER AS builder`
```
ARG GOWORK=off
ENV GOWORK=$GOWORK
```
So it's global to all the go commands we could run under the builder.
Then we would also need to pass that argument from the `Makefile`
```
docker-build: test ## Build docker image with the manager.
        podman build --build-arg GOWORK=$GOWORK -t ${IMG} .
```

Makefile
fmount: My only point here is that we can avoid `export GOWORK := $(GOWORK)` if [1] lands.
However, there isn't much change in how `govet.sh` and the other scripts are called, I guess the only thing we should take care is documenting you have to `GOWORK= make build` (before this change you didn't have to).
If we combine this change with [1] we can't just avoid the `export` directive (which defaults to `off` because of the CI execution) and keep the user experience as it is (and in line with the other operators).

[1] https://github.com/openstack-k8s-operators/openstack-k8s-operators-ci/pull/56

Akrog: I think everything in `Makefile` should not use the go workspace by default, That way build and run also ignores it unless explicitly requested with the `GOWORK=`.

I'm thinking about leftover `go.work` replaces after testing a PR and then forgetting about it and doing `make build` and getting all kinds of issues.

I believe it is less likely to forget adding the `GOWORK=` when we are working with dependencies than the other way around.

The [documentation PR already includes the `GOWORK=`](https://github.com/openstack-k8s-operators/cinder-operator/pull/77/commits/27f69cf3044bddbdb464acecef7c5b67d086a050) when using external dependencies.

Makefile
```diff
@@ -117,11 +120,17 @@ fmt: ## Run go fmt against code.
...
-	go vet ./...
+	go vet ./... ./api/...
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... ./api/... -coverprofile cover.out
+
+.PHONY: gowork
+gowork: ## Generate go.work file to support our multi module repository
```
Akrog, line 130: Do we still want this even though we don't use it anymore? (just checking, I don't have a strong opinion)

Makefile
fmount: > I think everything in `Makefile` should not use the go workspace by default, That way build and run also ignores it unless explicitly requested with the `GOWORK=`.
> 
> I'm thinking about leftover `go.work` replaces after testing a PR and then forgetting about it and doing `make build` and getting all kinds of issues.
> 
> I believe it is less likely to forget adding the `GOWORK=` when we are working with dependencies than the other way around.

Agreed, tbh I didn't have a strong opinion but as you mentioned, it's good making sure that people understand how to explicitly turn it on instead of messing with potential `go.work` changes (that would end up "working on the laptop").
Being that said, I added the `export` directive in the last iteration.
> 
> The [documentation PR already includes the `GOWORK=`](https://github.com/openstack-k8s-operators/cinder-operator/pull/77/commits/27f69cf3044bddbdb464acecef7c5b67d086a050) when using external dependencies.

Thank you, sounds good!

Makefile
fmount: Given it's an isolated target and we don't call it explicitly, for development purposes it's good to have it here in case you're fighting with the go dependencies model. I added it under "Development", so you can see it in the right column (see `make help`)

Akrog (approved):
Thanks for the PR and all the changes.
Code looks good to me

stuggi (approved):
/lgtm

PR #80, opened 2022-11-28
use stable/zed container images

SeanMooney:
This chnage makes the container image optional and
updates the defaults to use stable/zed images.


Akrog (approved):
/lgtm

PR #81, opened 2022-11-30
Makefile: Fix install and deploy

Akrog:
When installing CRDs in the cluster it will fail due to the length of the CRD.

  The CustomResourceDefinition "cinders.cinder.openstack.org" is
  invalid: metadata.annotations: Too long: must have at most 262144
  bytes

To fix this we'll remove the descriptions on the CRDs when running `install` and `deploy` targets.

This is done with `crd:maxDescLen=0` just like it has been done in the openstack-operator.

We didn't use to have this size problem, but now we do with the addition of the `extraVol`.


abays (approved):
/lgtm

PR #84, opened 2022-12-05
Add top level NodeSelector and wire into services

dprince:


abays (commented):
+1 from me

api/v1beta1/cinder_types.go
```diff
@@ -64,6 +64,10 @@ type CinderSpec struct {
 	// PasswordSelectors - Selectors to identify the DB and ServiceUser password from the Secret
 	PasswordSelectors PasswordSelector `json:"passwordSelectors,omitempty"`
 
+        // +kubebuilder:validation:Optional
+        // NodeSelector to target subset of worker nodes running Cinder services
+        NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
Akrog, line 69: Wrong indentation level.

Also, not sure we should only have the node selector at the top level, because volume and backup services are in the data path (eg: their nodes may need FC HBAs) that we may want to have them on separated nodes from API and scheduler that are hardware independent.

Akrog (changes requested):

Akrog:
I'm closing the PR since I now see that we have merged an equivalent PR that also has per-service NodeSelector: https://github.com/openstack-k8s-operators/cinder-operator/pull/95

PR #95, opened 2022-12-13
Add top level NodeSelector to Cinder CR

bshephar:
This change adds the top level NodeSelecter to the Cinder CR.  This can then be overridden by individual CR's as required.


abays (approved):
/lgtm

PR #99, opened 2022-12-19
Remove useless GetVolumes call

fmount:
When `dbsync` is called, both `Volumes` and `VolumeMounts` are defined in the `DbSyncJob` function (using the very same parameters). 
This change just removes the duplicated line that calls the same function.


abays (approved):
/lgtm

PR #103, opened 2023-01-10
Properly watch TransportURL secrets

abays:
We are currently watching `Secret`s in the Cinder controller, but we do not create any `Secret`s of our own, so it does nothing.  What we are trying to do is watch the derivative `Secret` created by the `TransportURL` that we create.  Using a watch function instead of `Owns()` will allow us to properly watch those derivative `Secret`s.

### DIFF
--- a/controllers/cinder_controller.go
+++ b/controllers/cinder_controller.go
@@ -27,7 +27,10 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/go-logr/logr"
 	cinderv1beta1 "github.com/openstack-k8s-operators/cinder-operator/api/v1beta1"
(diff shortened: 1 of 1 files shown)


controllers/cinder_controller.go
```diff
@@ -195,6 +198,40 @@ func (r *CinderReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
+			if ownerRef.Kind == "TransportURL" {
+				for _, cr := range cinders.Items {
+					if ownerRef.Name == fmt.Sprintf("%s-cinder-transport", cr.Name) {
+						// return namespace and Name of CR
+						name := client.ObjectKey{
+							Namespace: o.GetNamespace(),
+							Name:      cr.Name,
+						}
+						r.Log.Info(fmt.Sprintf("TransportURL Secret %s belongs to TransportURL belonging to Cinder CR %s", o.GetName(), cr.Name))
+						result = append(result, reconcile.Request{NamespacedName: name})
```
Akrog, line 224: I'm not sure this is the most efficient way of doing the check, but I'm probably missing something.

Couldn't we search the owner references first for the ownerRef.Name ending in "-cinder-transport" and only if found would we get the specific Cinder CR based on the name (if only 1 found) or the list (if more than 1 found)?

I think that would be faster for all non cinder transportURL related secrets.

Shouldn't we also monitor the other secrets we use?  eg: https://github.com/openstack-k8s-operators/cinder-operator/blob/master/api/v1beta1/cinderapi_types.go#L54

Akrog: I'm ok with a follow up for the other secrets, since it's a different issue from the one you are fixing here.

abays: Ack.  I will make the efficiency update here and then handle the other secrets in a separate PR.

abays: I looked at this again and remembered why I initially wrote this the "inefficient" way.  These watch functions return a list of namespace-scoped CRs that then get fed to the reconciler.  Hence we need to know the name [1] of the Cinder CR associated with the `secret` we are examining in the function.  We could parse the name out of the `"%s-cinder-transport"` `secret` label, which would be faster than getting the list of the Cinder CRs and trying to match on each one.  The only downside I see there, however, is that technically someone could randomly label a `secret` `something-cinder-transport` where `something` actually matches the name of an existing Cinder CR.  In that case changes to that `secret` would trigger reconciliation for a Cinder CR that does not need it.  I know it's highly unlikely -- and I suppose someone could just throw the whole label (`whatever-cinder-transport`) on a `secret` anyhow -- but I thought I'd just mention it.

Anyhow, @Akrog, if you think it's fine to just parse the Cinder CR name from the label itself, we can improve efficiency here by avoiding getting the list of Cinders and looping through them.

[1] https://github.com/abays/cinder-operator/blob/3e18f7927c678bc3acffcb7658dc5ad79079b7c2/controllers/cinder_controller.go#L219-L222

fmount: I think @Akrog is right from a pure efficiency point of view, we can watch/match
the `ownerRef.Name` that end with `cinder-transport`, then process the CRs.
However, I see that someone can just label `<whatever>-cinder-transport` and
trigger a reconciliation that isn't needed. In this case, the complexity of
executing the reconciliation is greater than a `for` loop over the CRs where
`owner.Name` ends with `cinder-transport`.
For this reason (and given the limited amount of transportURL secrets), I'm ok
to go with the "inefficient" way and make sure the code trigger the reconciliation
only if needed.
Not sure there's a better way of doing it, we should probably think more on how
this can be optimized, but we definitely don't want to trigger a reconciliation loop
over a secret which shouldn't be watched by cinder (this could be worst than a
missing optimization).
I think we can follow up on how other secrets should be watched [1], and in that
PR we can do more optimization.

[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/master/api/v1beta1/cinderapi_types.go#L54

fmount (approved):

controllers/cinder_controller.go
Akrog: We can move forward with this implementation, but I would appreciate if we could document the reasons mentioned in this conversation in the commit message or a comment in the function.

PR #106, opened 2023-01-16
Default CinderVolume and CinderBackup replicas to 0

fmount:
This patch changes the `CRD` default for both `CinderVolume` and `CinderBackup` replicas, setting it to `0`.
In addition, we need to make sure the `IsReady` returns `true` if no instances are defined.

### DIFF
--- a/api/v1beta1/cinderbackup_types.go
+++ b/api/v1beta1/cinderbackup_types.go
@@ -35,6 +35,7 @@ type CinderBackupSpec struct {
 	ContainerImage string `json:"containerImage,omitempty"`
 
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=0
 	// Replicas - Cinder Backup Replicas
 	Replicas int32 `json:"replicas"`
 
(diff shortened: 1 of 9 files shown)


abays:
We also need to update this section I think: https://github.com/openstack-k8s-operators/cinder-operator/blob/af806d8930fff1c4d3fe66f5e857728a915a3074/api/v1beta1/cinder_types.go#L191-L200

api/v1beta1/cinderbackup_types.go
```diff
@@ -132,5 +133,5 @@ func init() {
 
 // IsReady - returns true if service is ready to serve requests
 func (instance CinderBackup) IsReady() bool {
-	return instance.Status.ReadyCount >= 1
+	return instance.Status.ReadyCount >= 1 || instance.Spec.Replicas == 0
```
Akrog, line 136: This doesn't sound right to me.  Not only would the name be misleading with this change, but also the comment with the description of the method would be wrong, as we could return `true` when there is no service to serve requests.

fmount: You're right about this one, I wasn't sure since it could result in just hacking the IsReady() function. So here my question: do you think that we should update the controller (e.g. see [1] for cinder-volume) to not update the `Status` or do you think this PR doesn't make sense?

[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/master/controllers/cindervolume_controller.go#L120

PR #107, opened 2023-01-17
Use own user to send admin requests to nova

kajinamit:
... instead of nova user. This allows us to reduce dependency on the resources managed by nova-operator and better separate each operators. The credential should work fine as long as the user has admin access with the 'service' project scope.

### DIFF
--- a/templates/cinder/bin/init.sh
+++ b/templates/cinder/bin/init.sh
@@ -23,9 +23,7 @@ export DB=${DatabaseName:-"cinder"}
 export DBHOST=${DatabaseHost:?"Please specify a DatabaseHost variable."}
 export DBUSER=${DatabaseUser:-"cinder"}
 export DBPASSWORD=${DatabasePassword:?"Please specify a DatabasePassword variable."}
-export CINDERPASSWORD=${CinderPassword:?"Please specify a CinderPassword variable."}
-# TODO: nova password
-#export NOVAPASSWORD=${NovaPassword:?"Please specify a NovaPassword variable."}
+export PASSWORD=${CinderPassword:?"Please specify a CinderPassword variable."}
 export TRANSPORTURL=${TransportURL:-""}
 
 export CUSTOMCONF=${CustomConf:-""}
@@ -73,8 +71,7 @@ if [ -n "$TRANSPORTURL" ]; then
   crudini --set ${SVC_CFG_MERGED} DEFAULT transport_url $TRANSPORTURL
 fi
 crudini --set ${SVC_CFG_MERGED} database connection mysql+pymysql://${DBUSER}:${DBPASSWORD}@${DBHOST}/${DB}
-crudini --set ${SVC_CFG_MERGED} keystone_authtoken password $CINDERPASSWORD
-# TODO: nova password
-#crudini --set ${SVC_CFG_MERGED} nova password $NOVAPASSWORD
+crudini --set ${SVC_CFG_MERGED} keystone_authtoken password $PASSWORD
+crudini --set ${SVC_CFG_MERGED} nova password $PASSWORD
 # TODO: service token
 #crudini --set ${SVC_CFG_MERGED} service_user password $CinderPassword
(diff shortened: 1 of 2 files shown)


templates/cinder/bin/init.sh
```diff
@@ -73,8 +71,7 @@ if [ -n "$TRANSPORTURL" ]; then
   crudini --set ${SVC_CFG_MERGED} DEFAULT transport_url $TRANSPORTURL
 fi
 crudini --set ${SVC_CFG_MERGED} database connection mysql+pymysql://${DBUSER}:${DBPASSWORD}@${DBHOST}/${DB}
-crudini --set ${SVC_CFG_MERGED} keystone_authtoken password $CINDERPASSWORD
-# TODO: nova password
-#crudini --set ${SVC_CFG_MERGED} nova password $NOVAPASSWORD
+crudini --set ${SVC_CFG_MERGED} keystone_authtoken password $PASSWORD
+crudini --set ${SVC_CFG_MERGED} nova password $PASSWORD
```
bogdando, line 75: it is assumed to be a CinderPassword, but used for nova password as well. Would be nice to clarify the password equality requirements, or the like

kajinamit: Do you mind clarifying what you expect ? Do you want me to add a note (though the node has not yet been added to any other operators so I tend to avoid putting it only in cinder operator)

bogdando: having something noted about different passwords used would be nice, thanks

kajinamit: Just to avoid my misunderstanding. This implementation does not mean that we should use the same password for different service users.

As we are using the same cinder user, it's quite reasonable to use the same password. You can check the config file template to find out which user is used. I can add a note but if we add one then we have to add the same to every operator doing this. Do you want me to update all operators ? I see even nova operator does not contain such a note.

fmount: I think there's an effort to remove crudini from the init script (see #108): it's still in progress and it can be rebased on this change when it's ready.
Adding @ASBishop for awareness, but I don't see big problems rebasing his change on top of this.

fmount (approved):

templates/cinder/bin/init.sh
Akrog: We should probably document somewhere the expectations on the cinder user.

templates/cinder/bin/init.sh
```diff
@@ -23,9 +23,7 @@ export DB=${DatabaseName:-"cinder"}
 export DBHOST=${DatabaseHost:?"Please specify a DatabaseHost variable."}
 export DBUSER=${DatabaseUser:-"cinder"}
 export DBPASSWORD=${DatabasePassword:?"Please specify a DatabasePassword variable."}
-export CINDERPASSWORD=${CinderPassword:?"Please specify a CinderPassword variable."}
-# TODO: nova password
-#export NOVAPASSWORD=${NovaPassword:?"Please specify a NovaPassword variable."}
+export PASSWORD=${CinderPassword:?"Please specify a CinderPassword variable."}
```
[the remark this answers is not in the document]
kajinamit: Currently the same templates in the other service operators use PASSWORD, because the script uses the single password of its service user.
I can fix this but that makes this file only in cinder-operator looks different from the others. If we are fine with that "inconsistency" then I'll update it.

templates/cinder/bin/init.sh
kajinamit: We probably can discuss this in the next sync but I think we should document the general design (use own service user to interact with the other services) somewhere more common instead of adding comments in individual files.

Akrog: I didn't mean as a comment here, I was thinking more in terms of documentation.

templates/cinder/bin/init.sh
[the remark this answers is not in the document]
Akrog: Well, we will be changing/throwing away most of this file in the mid-term, so I'm ok being consistent with other projects for the time being.

Akrog (approved):
/lgtm

PR #115, opened 2023-01-20
Remove the old Ceph section from cinder.conf

fmount:
This patch just removes the unused ceph section in cinder.conf. 
The `extraVolumes` approach is used when cinder is configured with a Ceph backend, so this section is not required anymore.


fmount:
> Would you mind updating this PR to include updating the readme which still has cephBackend? Perhaps update it to have the new method. I could also take care of that in a follow up PR if you prefer to have two. Either way I agree we should have this change.
> 
> https://github.com/openstack-k8s-operators/cinder-operator/blob/7665a222fcfb6c869c202dc0805a3fd2f18dab70/README.md#configure-cinder-with-ceph-backend

I updated the Readme removing the old (and wrong) information. I added the reference to the `Ceph` secrets and `extraMounts`.
For the Readme, I think we should work on getting [1] merged to reflect the whole status of the operator.

[1] https://github.com/openstack-k8s-operators/cinder-operator/pull/77

stuggi (approved):
/lgtm

PR #123, opened 2023-01-31
Remove and ignore go.work

yazug:
Our build system does not support go workspaces[1]. So even though we have two modules in this repository we should not have a go.work file as that will fail the build. Still there are tools that only work properly in our repo if there is go.work file provided. For example go vet only checks the ./api module if the go.work file lists the ./api as the module of the workspace. So a gowork make target is added to generate the go.work file locally but never committed to the repository.

[1]https://github.com/containerbuildsystem/cachito/commit/c7fbbb13eeb7d7edec3c01643eb5b24381027365


Makefile
```diff
@@ -324,3 +325,8 @@ golint: get-ci-tools
 	PATH=$(GOBIN):$(PATH); $(CI_TOOLS_REPO_DIR)/test-runner/golint.sh
 	PATH=$(GOBIN):$(PATH); $(CI_TOOLS_REPO_DIR)/test-runner/golint.sh ./api
 
+.PHONY: gowork
```
SeanMooney, line 328: this is already defiend on line 131

https://github.com/openstack-k8s-operators/cinder-operator/pull/123/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R131

yazug: done

SeanMooney (commented):
This looks good to me but the cinder folks should review

Akrog (changes requested):
This probably doesn't work as intended for the cinder-operator repository.
Running `make vet` will fail with `go: no go.work file found` on the first `go work use .` invocation.
This happens because we are defaulting `GOWORK` to `off` on L44.

Akrog (changes requested):
I think we can do it how you were doing in the previous patch, where you added `vet: gowork`.

We just need to add `export GOWORK=` in `gowork` and `vet` targets to ensure they default to using workspaces:

```
vet: export GOWORK=
vet: gowork ## Run go vet against code.
	go vet ./... ./api/...

gowork: export GOWORK=
gowork: ## Generate go.work file to support our multi module repository
	test -f go.work || go work init
	go work use .
	go work use ./api
```

yazug:
I'm not sure how to test that what I am changing is doing the right thing.  Someone who can do the testing and works on this will need to help get any Makefile changes fixed up.

abays:
> I think we can do it how you were doing in the previous patch, where you added `vet: gowork`.
> 
> We just need to add `export GOWORK=` in `gowork` and `vet` targets to ensure they default to using workspaces:
> 
> ```
> vet: export GOWORK=
> vet: gowork ## Run go vet against code.
> 	go vet ./... ./api/...
> 
> gowork: export GOWORK=
> gowork: ## Generate go.work file to support our multi module repository
> 	test -f go.work || go work init
> 	go work use .
> 	go work use ./api
> ```

@yazug I tested these suggest changes locally and everything built fine, so they won't break dev work

fmount (approved):

fmount:
> > I think we can do it how you were doing in the previous patch, where you added `vet: gowork`.
> > We just need to add `export GOWORK=` in `gowork` and `vet` targets to ensure they default to using workspaces:
> > ```
> > vet: export GOWORK=
> > vet: gowork ## Run go vet against code.
> > 	go vet ./... ./api/...
> > 
> > gowork: export GOWORK=
> > gowork: ## Generate go.work file to support our multi module repository
> > 	test -f go.work || go work init
> > 	go work use .
> > 	go work use ./api
> > ```
> 
> @yazug I tested these suggest changes locally and everything built fine, so they won't break dev work

Same here, I quickly tested in my local env and it works as expected.

Akrog (approved):
/lgtm

PR #126, opened 2023-02-02
Remove go.work from repo

abays:
The `go.work` file is in the `.gitignore` and was seemingly mistakenly pushed into the repo


abays:
> @abays Have you seen this PR #123 some comments there from Gorka

@dprince Totally missed that.  Let's defer to that one.

PR #133, opened 2023-02-09
Use time.Duration for timeout

kajinamit:
instead of int. This allows us to use more fine grained timeout value such as 0.1, and encodes the unit in the type itself.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/155
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/166


gibizer (approved):
looks good

fmount (approved):

abays:
At least some of the CI failures are just plain wrong.  For instance:

```
internalEndpointData declared and not used (typecheck)
```

Not true at all: https://github.com/kajinamit/cinder-operator/blob/cb1d5184790b4b04928bcadd179b88dd4fe03952/controllers/cinderapi_controller.go#L310-L318

kajinamit:
Looks like I messed up go.mod somehow, which is fixed in the latest version. The failure in build-deploy does not look related to this change.

gibizer:
/test cinder-operator-build-deploy
I think it was cause by a missing ironic image in quay which was since fixed

gibizer (approved):
looks good

fmount (commented):
@kajinamit not sure I got the purpose of `binba/foo`

kajinamit:
> @kajinamit not sure I got the purpose of `binba/foo`

I have no idea how I added that file :-(

dprince (approved):

PR #136, opened 2023-02-14
Support attaching pods to networks using net-attach-defs and exposing api via metallb

stuggi:
* If needed, service pods can be attach to networks using the `networkAttachments` list in the CR. Its a list of pre-created net-attach-defs names. Per default the pods won't get attached.

```
spec:
...
      networkAttachments:
      - storage
```
Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/133

* introduce externalEndpoints Parameters for cinder api to configure metallb service using the pre-created metallb IPAddressPool and L2Advertisement . 

E.g. to expose the api on an internalapi network which
* should be registered as the internal endpoint
* its metallb ipaddresspool is configured in an osp-internalapi object
* and request a specific loadbalancerIP which is shared with other services (list of IPs for multi stack ipv4/ipv6)

```
...
spec:
...
      externalEndpoints:
      - endpoint: internal
        loadBalancerIPs:
        - 172.17.0.202
        ipAddressPool: osp-internalapi
        sharedIP: true
```

* removes registration of admin endpoint

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/160


api/v1beta1/common_types.go
```diff
@@ -54,3 +56,32 @@ type CinderServiceDebug struct {
 	// service enable debug
 	Service bool `json:"service,omitempty"`
 }
+
+// MetalLBConfig to configure the MetalLB loadbalancer service
+type MetalLBConfig struct {
```
fmount, line 61: I made a similar comment in the glance-operator patch and I'm wondering if we want to have a single definition of this MetalLB struct in lib-common instead of replicating it across the operators

stuggi: we plan to move this into the infra-operator when we add there more network related stuff, like ipam, dns and so to have it versioned there within the api. until then we keep it in the operators

fmount: Thanks, let's move to the `infra-operator` and import the corresponding struct in a follow up PR.

stuggi:
had to rebase

pkg/cinderbackup/statefulset.go
```diff
@@ -120,6 +121,8 @@ func StatefulSet(
 			Replicas: &instance.Spec.Replicas,
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
+					Annotations: annotations,
+
```
abays, line 125: Looks like an extra newline here

stuggi: ack, done

abays (approved):
/lgtm

PR #148, opened 2023-03-06
Webhook for default images

abays:
Default container images are now handled via env vars and webhooks, as per the agreed pattern we have recently initiated in Glance [1] and are integrating with OpenStack Operator via external (to the specific service operators) webhooks that use the new defaulting mechanism [2].

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/170
[2] https://github.com/openstack-k8s-operators/openstack-operator/pull/220


fmount (approved):
The code here seems in line with what we did in the `glance-operator`, so +1 from me.
I don't see particular issues blocking this patch, I'm wondering if we're able to test it with the meta operator and how we can do it locally.

PR #153, opened 2023-03-14
Remove description from Cinder CRD

fmount:
This patch rebuilds the Cinder `CRD` removing the description to avoid a potential issue when the cinder bundle is deployed via `OLM` by the meta operator and the following is seen:

```
openstack/install-7ws9s"} failed: failed to update installplan bundle lookups: etcdserver: request is too large
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -234,6 +234,7 @@ $(ENVTEST): $(LOCALBIN)
 	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
 .PHONY: bundle
+bundle: CRDDESC_OVERRIDE=:maxDescLen=0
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -234,6 +234,7 @@ $(ENVTEST): $(LOCALBIN)
 	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
 .PHONY: bundle
+bundle: CRDDESC_OVERRIDE=:maxDescLen=0
```
Akrog, line 237: I forgot to add this on https://github.com/openstack-k8s-operators/cinder-operator/commit/e33a589d51de900c82bde5843cf18d7c43cf421a , thanks for adding it.

Akrog (approved):
/lgtm

PR #155, opened 2023-03-14
Support local webhooks

abays:
Allows webhooks to run locally for dev/test via `make run-with-webhook`

### DIFF
--- a/hack/configure_local_webhook.sh
+++ b/hack/configure_local_webhook.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+set -ex
+
+TMPDIR=${TMPDIR:-"/tmp/k8s-webhook-server/serving-certs"}
+SKIP_CERT=${SKIP_CERT:-false}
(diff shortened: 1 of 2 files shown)


hack/configure_local_webhook.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
+set -ex
+
+TMPDIR=${TMPDIR:-"/tmp/k8s-webhook-server/serving-certs"}
+SKIP_CERT=${SKIP_CERT:-false}
+CRC_IP=${CRC_IP:-$(/sbin/ip -o -4 addr list crc | awk '{print $4}' | cut -d/ -f1)}
+
+#Open 9443
+sudo firewall-cmd --zone=libvirt --add-port=9443/tcp
```
fmount, line 9: in my case I don't have `firewall-cmd`, but long term we can do that or add the `iptables command`:

`sudo iptables -I INPUT -p tcp -m tcp --dport 9443 -j ACCEPT  && service iptables save`

anyway looks good as long as that's the default used.

PR #157, opened 2023-03-14
Add CRDDESC_OVERRIDE to 'manifests' Makefile target

ASBishop:


ASBishop:
Not needed, see https://github.com/openstack-k8s-operators/cinder-operator/pull/156

PR #159, opened 2023-03-24
Allow cinderbackup to report ready with 0 replicas

stuggi:
Right now the cinder-backup resource gets created with default 0 replicas. With this cinder never reaches ready state without cinder-backup being configured. This change reports Ready state if `Spec.Replicas == ReadyCount`.

### DIFF
--- a/api/v1beta1/cinderbackup_types.go
+++ b/api/v1beta1/cinderbackup_types.go
@@ -146,5 +146,5 @@ func init() {
 
 // IsReady - returns true if service is ready to serve requests
 func (instance CinderBackup) IsReady() bool {
-	return instance.Status.ReadyCount >= 1
+	return instance.Status.ReadyCount == instance.Spec.Replicas
 }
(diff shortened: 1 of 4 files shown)


stuggi:
@Akrog @fmount @fultonj wondering if it is expected that cinder-backup is always configured or if we want to have cinder-backup to report back that its Ready with set 0 replicas

api/v1beta1/cinderbackup_types.go
```diff
@@ -146,5 +146,5 @@ func init() {
 
 // IsReady - returns true if service is ready to serve requests
 func (instance CinderBackup) IsReady() bool {
-	return instance.Status.ReadyCount >= 1
+	return instance.Status.ReadyCount >= 1 || instance.Spec.Replicas == 0
```
Akrog, line 149: Should we change this to be?:

```
  return (instance.Status.ReadyCount >= 1 && instance.Spec.Replicas > 0) || (instance.Spec.Replicas == instance.Status.ReadyCount)
```

Because I think it's wrong to say we are ready when we change from replica 1 to replica 0 when we still have the replica up and running, right?

stuggi: will update the pr as discussed on slack

Akrog (approved):
/lgtm

PR #160, opened 2023-03-24
Set component labels on deployment resources

stuggi:
Network validation is checking the pods if IPs are correct reported for each of the requested networks. For this a label selecor is used to get only the pods for the sub cinder component (api,scheduler, backup and volume). Right now there is only the generic cinder app label on the deployment resources set.
This adds a component label on all the api, scheduler, backup and volume deployment resources which allows the check to only get the wanted resource pods.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/225


stuggi:
needs go.mod update when https://github.com/openstack-k8s-operators/lib-common/pull/225 got merged

abays (approved):
/lgtm

PR #161, opened 2023-03-24
Get overall state from sub resources Ready state

stuggi:


api/v1beta1/cinder_types.go
```diff
@@ -186,16 +186,12 @@ func init() {
...
-	ready := instance.Status.CinderAPIReadyCount > 0 && instance.Status.CinderBackupReadyCount > 0 &&
-		instance.Status.CinderSchedulerReadyCount > 0
-
-	for name := range instance.Spec.CinderVolumes {
-		ready = ready && instance.Status.CinderVolumesReadyCounts[name] > 0
-	}
-
-	return ready
+	return instance.Status.Conditions.IsTrue(CinderAPIReadyCondition) &&
+		instance.Status.Conditions.IsTrue(CinderBackupReadyCondition) &&
```
ASBishop, line 192: We'll have to be careful about determining the value of these various ready conditions. The rules for determining "how ready is ready" are likely to be unique. If I have >0 c-api and c-sch then that's probably OK. And if there are zero c-bak (it's an optional service) then that should be OK. The c-bak service is ready because it's not part of the deployment! As for the c-vol services, is it ready if any of them are ready? What happens if one backend is down? That should not imply that all of cinder is not ready.

ASBishop: Ah, it's clear to me now, thanks!

Akrog (approved):
/lgtm

PR #162, opened 2023-03-24
Backend config samples

Akrog:
This patch adds a Ceph backend sample and creates the skeleton for other backend samples.

The idea is for people to be able to just look at one of the files in `config/samples/backends/*/backend.yaml` and see just the relevant bits to configure the Cinder backend.

Other files may be added in there, since credentials will be provided in a different manifest and then referenced in the `backend.yaml` file using the `customServiceConfigSecrets`


config/samples/backends/bases/openstack.yaml
```diff
@@ -0,0 +1,139 @@
...
+        replicas: 1
+        containerImage: quay.io/tripleozedcentos9/openstack-cinder-api:current-tripleo
+      cinderScheduler:
+        replicas: 1
+        containerImage: quay.io/tripleozedcentos9/openstack-cinder-scheduler:current-tripleo
+      cinderBackup:
+        replicas: 0
+        containerImage: quay.io/tripleozedcentos9/openstack-cinder-backup:current-tripleo
+      cinderVolumes:
+        volume1:
```
ASBishop, line 50: I realize this is a generic baseline CR, but I wish we could move away from using the "volume1" name, and defer until a specific backend is deployed.

config/samples/backends/ceph/backend.yaml
```diff
@@ -0,0 +1,61 @@
...
+              - secret:
+                  name: ceph-conf-files
+          mounts:
+          - name: ceph
+            mountPath: "/etc/ceph"
+            readOnly: true
+  cinder:
+    template:
+      cinderVolumes:
+        volume1:
```
ASBishop, line 30: Per my previous comment, it would be nice if this c-vol could be named "ceph" instead of "volume1"

config/samples/backends/ceph/backend.yaml
```diff
@@ -0,0 +1,61 @@
...
+          customServiceConfig: |
+            [ceph]
+            volume_backend_name=ceph
+            volume_driver=cinder.volume.drivers.rbd.RBDDriver
+            rbd_ceph_conf=/etc/ceph/ceph.conf
+            rbd_user=openstack
+            rbd_pool=volumes
+            rbd_flatten_volume_from_snapshot=False
+            report_discard_supported=True
+            backend_host=hostgroup
```
ASBishop, line 41: I don't think we want this (no pcmk!)

config/samples/backends/bases/openstack.yaml
```diff
@@ -0,0 +1,139 @@
...
+  glance:
+    template:
+      databaseInstance: openstack
+      containerImage: quay.io/tripleozedcentos9/openstack-glance-api:current-tripleo
+      storageClass: ""
+      storageRequest: 2G
+      glanceAPIInternal:
+        containerImage: quay.io/tripleozedcentos9/openstack-glance-api:current-tripleo
+      glanceAPIExternal:
+        containerImage: quay.io/tripleozedcentos9/openstack-glance-api:current-tripleo
```
ASBishop, line 34: In my test deployment, I eliminated the containerImage lines, but fould that I had to replace them with "replicas: 1" to satisfy a requirement to have something present for the glanceAPIInternal and glanceAPIExternal sections.

config/samples/backends/bases/openstack.yaml
ASBishop: Update: In my local tests, I was able to eliminate the cinderVolumes section entirely! (L49-L52). This made it possible to defer naming the c-vol pod(s) until the backends are specified.

config/samples/backends/ceph/backend.yaml
ASBishop: Update: Per my update above, by eliminating the entire cinderVolumes section in the base CR, I was able to choose a name  like "ceph" here, when the c-vol pod is added.

config/samples/backends/ceph/backend.yaml
```diff
@@ -0,0 +1,61 @@
...
+      cinderBackup:
+        replicas: 1
+        customServiceConfig: |
+          [DEFAULT]
+          debug = true
```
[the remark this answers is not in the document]
Akrog: Thanks. I initially had it on the volume and backup service, moved it to the base YAML and removed it from the volume service, but forgot to remove it from here.

config/samples/backends/ceph/backend.yaml
Akrog: That's right, we don't need it and we don't want it, but it's not because we don't have peacemaker but because we use a `StatefulSet` for the Volume service, because it takes care of always having the same hostname regardless of the node where it is running.  If that was not the case then we would need to set the `backend_host` on the operator ourselves.

config/samples/backends/bases/openstack.yaml
Akrog: Setting only the replicas instead of the image will make it more future-proof.

config/samples/backends/ceph/backend.yaml
Akrog: I was doing it for the same reason as above. Changing it now.

config/samples/backends/ceph/backend.yaml
```diff
@@ -0,0 +1,60 @@
...
+                  name: ceph-conf-files
+          mounts:
+          - name: ceph
+            mountPath: "/etc/ceph"
+            readOnly: true
+  cinder:
+    template:
+      cinderVolumes:
+        ceph:
+          replicas: 1
```
ASBishop, line 32: Do we want this to be explicit? It could be eliminated by relying on the default value being 1.

config/samples/backends/ceph/backend.yaml
```diff
@@ -0,0 +1,60 @@
...
+            [ceph]
+            volume_backend_name=ceph
+            volume_driver=cinder.volume.drivers.rbd.RBDDriver
+            rbd_ceph_conf=/etc/ceph/ceph.conf
+            rbd_user=openstack
+            rbd_pool=volumes
+            rbd_flatten_volume_from_snapshot=False
+            report_discard_supported=True
+      cinderBackup:
+        replicas: 1
```
ASBishop, line 43: Ack, here we do need to specify the number of replicas.

config/samples/backends/ceph/backend.yaml
```diff
@@ -0,0 +1,60 @@
...
+          - name: ceph
+            mountPath: "/etc/ceph"
+            readOnly: true
+  cinder:
+    template:
+      cinderVolumes:
+        ceph:
+          replicas: 1
+          customServiceConfig: |
+            [ceph]
```
ASBishop, line 34: Unless/until [1] merges, with zed container images it will be necessary to specify the rbd_secret_uuid (the cluster's FSID) for nova. As this is just a sample file, should it include a comment that it's the user's responsibility to add that?

[1] https://review.opendev.org/c/openstack/cinder/+/878168

ASBishop (commented):
Just a couple of questions, otherwise looks good!

config/samples/backends/ceph/backend.yaml
Akrog: I was planning to wait until we have nova working with cinder and then update the example to use Antelope containers.

config/samples/backends/ceph/backend.yaml
Akrog: I like it explicit, but just because it makes it easier to play with it.
But since that's not the objective of the example we should remove it to leave only what's necessary.

ASBishop (approved):
/lgtm

PR #167, opened 2023-03-27
Fix iscsi lvm

Akrog:
Fix iSCSI and multipathd on cinder volume and backup pods.
To allow testing it we also fix the LVM driver and provide a simple mechanism to run this backend on CRC.

In OpenShift the approach is to have iscsid and multipathd running on the host and using `nsenter` inside the container to communicate with them.

Commit messages have additional information.

### DIFF
--- a/config/samples/backends/lvm-iscsi/backend.yaml
+++ b/config/samples/backends/lvm-iscsi/backend.yaml
@@ -1,0 +1,25 @@
+# This sample requires an LVM VG with the name of `cinder-volumes` to exist on the OpenShift node where cinder-volume runs
+# It also requires that iscsid and multpathd are running on the host.
+# The correct way to configure and run the daemons is using `MachineConfig` as shown in `iscsid.yaml` and `multipathd.yaml`.
+# To create the LVM on CRC the sample also uses a `MachineConfig` using a systemd unit with file `lvm.yaml`
+#
+# ATTENTION: After applying this OpenShift nodes will reboot, because of the `MachineConfig` changes and will take a while to recover.
+
+apiVersion: core.openstack.org/v1beta1
+kind: OpenStackControlPlane
+metadata:
+  name: openstack
+spec:
+  cinder:
+    template:
+      cinderVolumes:
+        lvm-iscsi:
+          customServiceConfig: |
+            [lvm]
+            image_volume_cache_enabled=false
+            use_multipath_for_image_xfer=true
+            volume_driver=cinder.volume.drivers.lvm.LVMVolumeDriver
+            volume_group=cinder-volumes
+            target_protocol=iscsi
+            target_helper=lioadm
+            volume_backend_name=lvm_iscsi
--- a/config/samples/backends/lvm-iscsi/lvm.yaml
+++ b/config/samples/backends/lvm-iscsi/lvm.yaml
@@ -1,0 +1,35 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: master
+    service: cinder
+    component: cinder-volume
+  name: 99-master-cinder-lvm-losetup
+spec:
+  config:
+    ignition:
+      version:  3.2.0
+    Systemd:
+      Units:
+      - name: cinder-lvm-losetup.service
+        enabled: true
+        Contents:  |
+          [Unit]
+          Description=Cinder LVM losetup
+          DefaultDependencies=no
+          Conflicts=umount.target
+          Requires=lvm2-monitor.service systemd-udev-settle.service
+          Before=local-fs.target umount.target
+          After=var.mount lvm2-monitor.service systemd-udev-settle.service
+
+          [Service]
+          Environment=VG_GB_SIZE=10G
+          Environment=LOOPBACK_FILE=/var/home/core/cinder-volumes
+          Type=oneshot
+          ExecStart=bash -c "if [[ ! -e ${LOOPBACK_FILE} ]]; then /bin/truncate -s ${VG_GB_SIZE} ${LOOPBACK_FILE} && /sbin/vgcreate cinder-volumes `/sbin/losetup --show -f ${LOOPBACK_FILE}`; else /sbin/losetup -f ${LOOPBACK_FILE}; fi"
+          ExecStop=bash -c "/sbin/losetup -d `/sbin/losetup -j ${LOOPBACK_FILE} -l -n -O NAME`"
+          RemainAfterExit=yes
+
+          [Install]
+          WantedBy=local-fs-pre.target
--- a/config/samples/backends/lvm-iscsi/multipathd.yaml
+++ b/config/samples/backends/lvm-iscsi/multipathd.yaml
@@ -1,0 +1,30 @@
+# Includes the /etc/multipathd.conf contents and the systemd unit changes
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: master
+    service: cinder
+  name: 99-master-cinder-enable-multipathd
+spec:
+  config:
+    ignition:
+      version: 3.2.0
+    storage:
+      files:
+        - path: /etc/multipath.conf
+          overwrite: false
+          # Mode must be decimal, this is 0600
+          mode: 384
+          user:
+            name: root
+          group:
+            name: root
+          contents:
+            # Source can be a http, https, tftp, s3, gs, or data as defined in rfc2397.
+            # This is the rfc2397 text/plain string format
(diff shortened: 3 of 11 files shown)


Akrog:
I'm going to look into how we can tie together the control plane services with the enablement of the services on the host, because they need to be deployed using `MachineConfig`.
If tying it together should be done by the Human Operator and not the `cinder-operator`, then I will add an example to replace the `crc-start-iscsi.sh` script.

templates/cinder/bin/run-outside
ASBishop, line 1: Is "run-outside" an existing convention for this script? The name feels a little strange to me (Outside? Outside what?). Something like "run-on-host" is less vague (at least for me).

Akrog: Sounds good, I'll make the change.

templates/cinder/bin/run-outside
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec nsenter -a -t 1 -- `realpath -s $0` "$@"
```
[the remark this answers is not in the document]
Akrog: Yes, `realpath` is important, because the $PATH could be different in the host and the container, so a binary that could be found in the container via $PATH may not be found on the host.   Using `realpath` resolves that issue for us.

I wrote the `run-outside` script 2 years ago [1], and to be honest I don't recall the reason behind enclosing it in double quotes.

[1]: https://github.com/embercsi/ember-csi/commit/bda33682e73b5299fa176908691272d5751a7377

config/samples/backends/lvm-iscsi/backend.yaml
```diff
@@ -0,0 +1,26 @@
...
+  name: openstack
+spec:
+  cinder:
+    template:
+      cinderVolumes:
+        lvm-iscsi:
+          customServiceConfig: |
+            [lvm]
+            image_volume_cache_enabled=false
+            use_multipath_for_image_xfer=true
```
ASBishop, line 21: I have some enhancements to the default cinder.conf in mind, and would add this to the [backend_defaults] section. My understanding is there's no reason to _not_ set it True, so might as well make it the default setting for every backend that might use it.

config/samples/backends/lvm-iscsi/backend.yaml
```diff
@@ -0,0 +1,26 @@
...
+            volume_driver=cinder.volume.drivers.lvm.LVMVolumeDriver
+            volume_group=cinder-volumes
```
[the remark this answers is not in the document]
Akrog: Yeah, but I like making things explicit  ;-)

config/samples/backends/lvm-iscsi/multipathd.yaml
```diff
@@ -0,0 +1,29 @@
...
+  name: 99-master-custom-enable-multipathd
+spec:
+  config:
+    ignition:
+      version: 3.2.0
+    storage:
+      files:
+        - path: /etc/multipath.conf
+          overwrite: false
+          # Mode must be decimal, this is 0600
```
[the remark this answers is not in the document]
Akrog: I had the same thought

config/samples/backends/lvm-iscsi/multipathd.yaml
```diff
@@ -0,0 +1,29 @@
...
+    machineconfiguration.openshift.io/role: master
+  name: 99-master-custom-enable-multipathd
+spec:
+  config:
+    ignition:
+      version: 3.2.0
+    storage:
+      files:
+        - path: /etc/multipath.conf
+          overwrite: false
```
ASBishop, line 15: I'm not sure how this is supposed to work. This MachineConfig seems to just define the multipath.conf file, but doesn't the service need to be started? Does it need to be installed?

Akrog: You can see how we start it on L26-L29 of this same file.
I mention it in the comment of the first line.

hack/dev/crc-create-lvm.sh
```diff
@@ -0,0 +1,11 @@
...
+set -ev
+set -x
+
+LOCATION=$(realpath `dirname -- $BASH_SOURCE[0]`)
+source "$LOCATION/helpers.sh"
+
+loopback_file="/var/home/core/cinder-volumes"
+echo Creating $loopback_file inside the CRC VM
+crc_ssh "if [[ ! -e $loopback_file ]]; then truncate -s 10G $loopback_file; fi"
+crc_ssh "if ! crc_ssh sudo vgdisplay cinder-volumes; then sudo vgcreate cinder-volumes \`sudo losetup --show -f $loopback_file\` && sudo vgscan; fi"
```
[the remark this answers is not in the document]
Akrog: I thought about it, and decided to be lazy and use what I had.
But now that you point out my laziness for all the world to see, I guess I'll do it. ;-)

config/samples/backends/lvm-iscsi/multipathd.yaml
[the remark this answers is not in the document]
ASBishop: I'm seeing examples [1],[2] that suggest it recognizes octal.

[1] https://docs.openshift.com/container-platform/4.8/post_installation_configuration/machine-configuration-tasks.html
[2] https://cloud.redhat.com/blog/red-hat-enterprise-linux-coreos-customization

config/samples/backends/lvm-iscsi/multipathd.yaml
```diff
@@ -0,0 +1,30 @@
...
+          # Mode must be decimal, this is 0600
+          mode: 384
+          user:
+            name: root
+          group:
+            name: root
+          contents:
+            # Source can be a http, https, tftp, s3, gs, or data as defined in rfc2397.
+            # This is the rfc2397 text/plain string format
+            source: data:,defaults%20%7B%0A%20%20user_friendly_names%20no%0A%20%20skip_kpartx%20yes%0A%20%20find_multipaths%20yes%0A%7D%0A%0Ablacklist%20%7B%0A%7D
```
ASBishop, line 26: The same two (see above) links show an example of using "inline" instead of "source". That would look like this:

```
          contents:
            inline: |
              defaults {
                user_friendly_names no
                skip_kpartx yes
                find_multipaaths yes
              }

              blacklist {
              }


```

config/samples/backends/lvm-iscsi/multipathd.yaml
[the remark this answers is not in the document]
Akrog: If you look at those examples, those are not OpenShift `MachineConfig` themselves.
They are butane files, which require the butane tool which is not installed by default, so it requires [users to install the butane tool](https://docs.openshift.com/container-platform/4.8/installing/install_config/installing-customizing.html#installation-special-config-butane_installing-customizing).
That would mean that the example is not usable by default on a normal deployment.

config/samples/backends/lvm-iscsi/lvm.yaml
```diff
@@ -0,0 +1,33 @@
...
+          DefaultDependencies=no
+          Conflicts=umount.target
+          Requires=lvm2-monitor.service systemd-udev-settle.service
+          Before=local-fs.target umount.target
+          After=var.mount lvm2-monitor.service systemd-udev-settle.service
+
+          [Service]
+          Environment=LOOPBACK_FILE=/var/home/core/cinder-volumes
+          Type=oneshot
+          ExecStart=bash -c "if [[ ! -e ${LOOPBACK_FILE} ]]; then /bin/truncate -s 10G ${LOOPBACK_FILE} && /sbin/vgcreate cinder-volumes `/sbin/losetup --show -f ${LOOPBACK_FILE}`; else /sbin/losetup -f ${LOOPBACK_FILE}; fi"
```
ASBishop, line 28: What would it take to make the 10G size configurable?

config/samples/backends/lvm-iscsi/multipathd.yaml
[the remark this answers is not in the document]
ASBishop: Argh, you're right! I just read the ignition spec [1], and it confirms all the insanity we noted is correct (decimal mode, no support for inline content).

Update: you also supplied this link in the next comment's response.

[1] https://coreos.github.io/ignition/configuration-v3_4/

config/samples/backends/lvm-iscsi/multipathd.yaml
ASBishop: Unfortunately, this is true >_<

ASBishop:
You also might want to add labels (e.g. service: cinder, maybe component: cinder-volume) to the new CRDs

config/samples/backends/lvm-iscsi/lvm.yaml
Akrog: I don't think that's possible with kustomize, and I feel like we are getting sidetracked.  This is just a toy system (we are using a loop device, not even a real device) to show how it can be used.
I'll make it easier to see the size by adding the `10G` as an `Environment`

ASBishop: Yeah, let's leave this along for now. I know that in tripleo the size is configurable because they need to bump it up to run tempest in some ci jobs, but that's because they have jobs that use the lvm backend.

ASBishop (approved):
/lgtm

PR #169, opened 2023-03-28
Add NFS backend config sample

ASBishop:
This config sample shows how to deploy an NFS backend, with sensitive settings (the nas_host and nas_share_path) stored in a Secret referenced in the backend's customServiceConfigSecrets.


Akrog (approved):
/lgtm

PR #170, opened 2023-03-29
Clean up cinder-volume deployments

ASBishop:
When an entry in the CinderVolumes spec is deleted, the corresponding cinder-volume deployment (i.e. the c-vol backend) must also be deleted. This is done by scanning the list of cinder-volume deployments, and deleting them if their name doesn't appear in the CinderVolumes spec.

Note that removing the cinder-volume backend currently does not clean up the service entry in the cinder DB. That will be handled in a follow-up patch.


Akrog (approved):
/lgtm

PR #171, opened 2023-03-30
Make deploying cinder-backup optional

ASBishop:
Many OpenStack deployments do not use the cinder-backup service. With this in mind, the service should not be deployed unless the number of replicas is greater than zero. Conversely, the deployment should be cleaned up (deleted) if the service is scaled down to zero replicas.

### DIFF
--- a/config/samples/backends/bases/openstack/openstack.yaml
+++ b/config/samples/backends/bases/openstack/openstack.yaml
@@ -38,7 +38,6 @@ spec:
         replicas: 1
       cinderScheduler:
         replicas: 1
-      # Can omit the cinderBackup section. Making it explicit for reference.
       cinderBackup:
         replicas: 0
   manila:
--- a/controllers/cinder_controller.go
+++ b/controllers/cinder_controller.go
@@ -640,7 +640,6 @@ func (r *CinderReconciler) reconcileNormal(ctx context.Context, instance *cinder
 		instance.Status.Conditions.Set(c)
 	}
 
-	// TODO: These will not work without rabbit yet
 	// deploy cinder-scheduler
 	cinderScheduler, op, err := r.schedulerDeploymentCreateOrUpdate(ctx, instance)
 	if err != nil {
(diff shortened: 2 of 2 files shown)


controllers/cinder_controller.go
```diff
@@ -630,31 +629,50 @@ func (r *CinderReconciler) reconcileNormal(ctx context.Context, instance *cinder
...
+	} else {
+		// Clean up cinder-backup if there are no replicas
+		err = r.backupCleanupDeployment(ctx, instance)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	if backupCondition != nil {
+		// If there was a Status=False condition, set that as the CinderBackupReadyCondition
```
abays, line 667: It's not necessarily the case that the condition here is `Status=False`.  Maybe we should change the comment to remove that part?

abays (commented):
+1 from me

config/samples/backends/bases/openstack/openstack.yaml
```diff
@@ -38,7 +38,6 @@ spec:
         replicas: 1
       cinderScheduler:
         replicas: 1
-      # Can omit the cinderBackup section. Making it explicit for reference.
```
Akrog, line 41: Don't know why you are removing this comment.

controllers/cinder_controller.go
```diff
@@ -945,6 +963,26 @@ func (r *CinderReconciler) backupDeploymentCreateOrUpdate(ctx context.Context, i
...
+func (r *CinderReconciler) backupCleanupDeployment(ctx context.Context, instance *cinderv1beta1.Cinder) error {
+	deployment := &cinderv1beta1.CinderBackup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-backup", instance.Name),
+			Namespace: instance.Namespace,
+		},
+	}
+	key := client.ObjectKeyFromObject(deployment)
+	err := r.Client.Get(ctx, key, deployment)
+	if err == nil && cinder.GetOwningCinderName(deployment) == instance.Name {
```
Akrog, line 975: - I think we should log a warning if we find the object but we don't own it.
- Shouldn't we check that the error is not found and not a different error?

controllers/cinder_controller.go
```diff
@@ -630,31 +629,50 @@ func (r *CinderReconciler) reconcileNormal(ctx context.Context, instance *cinder
...
+	}
+
+	if backupCondition != nil {
+		// If there's a backupCondition then set that as the CinderBackupReadyCondition
+		instance.Status.Conditions.Set(backupCondition)
+	} else {
+		// The CinderBackup is ready, even if the service wasn't deployed.
+		// Using "condition.DeploymentReadyMessage" here because that is what gets mirrored
+		// as the message for the other Cinder children when they are successfully-deployed
+		instance.Status.Conditions.MarkTrue(cinderv1beta1.CinderBackupReadyCondition, condition.DeploymentReadyMessage)
```
Akrog, line 673: On initial deployment we can set it like this because there's nothing running, but when modifying a system that has cinder-backup present we need to wait until the CinderBackup has been actually removed, which should only happen once the statefulset and pods have actually been removed, right?

Akrog (changes requested):

controllers/cinder_controller.go
ASBishop: I can definitely expand things to add addtional log message for various error conditions. The two that come to mind are:

1. Any error other than "not found"
2. The deployment isn't owned by us

controllers/cinder_controller.go
ASBishop: My understanding is the CinderXxxReadyCondition(s) are used for the purpose of declaring the cinder service itself is ready, and generally that means the number of "ready" replicas is either >0 when replicas is 0, or when replicas *is* 0. For that reason I don't see when it's necessary to wait for the final backup replica to be deleted before declaring the service is "ready"

controllers/cinder_controller.go
ASBishop: Actually, I think the whole "not owned by us" it extraneous when the team agreed there would be only one cinder CR per namespace. I'm going to push an update that updates the logging, but it will also use lib-common code to look for a backup deployment in the given namespace.

Akrog (approved):
/lgtm

PR #172, opened 2023-03-31
Add service and component labels to more objects

ASBishop:
Add the service and component labels to the c-api Deployment, and c-bak, c-sch and c-vol StatefulSets, and to all ConfigMaps.

This significantly improves the ability to search for all resources owned by cinder, or one of its components:

  oc get all -l service=cinder
  oc get all -l component=cinder-volume

"oc get all" doesn't list Secrets or ConfigMaps, but these can also be queried.

  oc get secret -l service=cinder
  oc get cm -l component=cinder-volume


Akrog (approved):
/lgtm

PR #174, opened 2023-03-31
Enhancements to default cinder.conf

ASBishop:
- Add [backend_defaults] and set use_multipath_for_image_xfer True
- Configure the [service_user] section
- Minor cleanup: removed obsolete comments, and sort the sections alphabetically. The sorting isn't critcal, but I didn't like seeing the oslo sections scattered all over the place.

### DIFF
--- a/templates/cinder/config/cinder.conf
+++ b/templates/cinder/config/cinder.conf
@@ -1,5 +1,4 @@
 [DEFAULT]
-# transport_url = rabbit://openstack:RABBIT_PASS@controller
 auth_strategy = keystone
 # TODO: Decide if we want the operator to generate the api servers, which is
 #       more efficient when creating volumes from image (no keystone requests).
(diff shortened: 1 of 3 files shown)


templates/cinder/config/cinder.conf
```diff
@@ -65,19 +72,11 @@ user_domain_name = Default
...
-#username = cinder
-#user_domain_name = Default
-#auth_type = password
-
-[oslo_concurrency]
-lock_path = /var/locks/openstack/cinder
-
-[os_brick]
-lock_path = /var/locks/openstack/os-brick
+[service_user]
```
Akrog, line 75: nit: it would have been good to mention that we are setting the password in `init.sh`.

Akrog: Then we need to find a way to document this for developers.

Akrog (approved):
/lgtm

PR #178, opened 2023-04-03
Use consistent words for keywords

kajinamit:
This updates the keywords of cinder CRD so that the items are more consistent with the other operators.

### DIFF
--- a/config/manifests/bases/cinder-operator.clusterserviceversion.yaml
+++ b/config/manifests/bases/cinder-operator.clusterserviceversion.yaml
@@ -54,8 +54,9 @@ spec:
   - supported: true
     type: AllNamespaces
   keywords:
-  - openstack
-  - cn-openstack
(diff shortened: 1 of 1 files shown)


config/manifests/bases/cinder-operator.clusterserviceversion.yaml
```diff
@@ -54,8 +54,9 @@ spec:
   - supported: true
     type: AllNamespaces
   keywords:
-  - openstack
-  - cn-openstack
+  - OpenStack
+  - Volume
```
Akrog, line 58: Wouldn't it be better to use `Block Storage` instead of `Volume`?

kajinamit: Ah, yes. I was referring to service type but Block Storage does sound better.

PR #180, opened 2023-04-03
Add label to TransportURL

Akrog:
We recently added labels to ConfigMaps and other service resources, but we didn't add them to the `TransportURL`.

This patch adds the `service=cinder` label to the `TransportURL` that is missing it.


ASBishop (approved):
/lgtm

PR #181, opened 2023-04-03
Fix NVMe-oF

Akrog:
Fix NVMe-oF on cinder volume and backup pods.
It is important to load the `nvme-fabrics` kernel module on all the nodes that can run those 2 services.
It also adds a sample to use NVMe-TCP with LVM volumes similar to what we do with iSCSI.

### DIFF
--- a/config/samples/backends/lvm-nvme-tcp/backend.yaml
+++ b/config/samples/backends/lvm-nvme-tcp/backend.yaml
@@ -1,0 +1,27 @@
+# This sample requires an LVM VG with the name of `cinder-volumes` to exist on the OpenShift node where cinder-volume runs
+# It also requires that nvme-fabrics kernel module is running everywhere and that nvmet is running on the controller host
+# The correct way to configure the load of additional kernel modules is using `MachineConfig` as shown in `nvme-fabrics.yaml`.yaml`.
+# To create the LVM on CRC the sample also uses a `MachineConfig` using a systemd unit with file `lvm.yaml`
+#
+# ATTENTION: After applying this OpenShift nodes will reboot, because of the `MachineConfig` changes and will take a while to recover.
+apiVersion: core.openstack.org/v1beta1
+kind: OpenStackControlPlane
+metadata:
+  name: openstack
+spec:
+  cinder:
+    template:
+      cinderVolumes:
+        lvm-nvme-tcp:
+          customServiceConfig: |
+            [lvm]
+            volume_backend_name=lvm_nvme_tcp
+            volume_driver=cinder.volume.drivers.lvm.LVMVolumeDriver
+            target_helper=nvmet
+            target_protocol=nvmet_tcp
+            target_port=4420
+            target_prefix=nvme-subsystem-1
+            nvmeof_conn_info_version=2
+            use_multipath_for_image_xfer=false
+            volume_group=cinder-volumes
+
--- a/config/samples/backends/lvm-nvme-tcp/lvm.yaml
+++ b/config/samples/backends/lvm-nvme-tcp/lvm.yaml
@@ -1,0 +1,36 @@
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    machineconfiguration.openshift.io/role: master
+    service: cinder
+    component: cinder-volume
+  name: 99-master-cinder-lvm-losetup
+spec:
+  config:
+    ignition:
+      version:  3.2.0
+    Systemd:
+      Units:
+      - name: cinder-lvm-losetup.service
+        enabled: true
+        Contents:  |
+          [Unit]
+          Description=Cinder LVM losetup
+          DefaultDependencies=no
+          Conflicts=umount.target
+          Requires=lvm2-monitor.service systemd-udev-settle.service
+          Before=local-fs.target umount.target
+          After=var.mount lvm2-monitor.service systemd-udev-settle.service
+
+          [Service]
+          Environment=VG_GB_SIZE=10G
+          Environment=LOOPBACK_FILE=/var/home/core/cinder-volumes
+          Type=oneshot
+          ExecStart=bash -c "if [[ ! -e ${LOOPBACK_FILE} ]]; then /bin/truncate -s ${VG_GB_SIZE} ${LOOPBACK_FILE} && /sbin/vgcreate cinder-volumes `/sbin/losetup --show -f ${LOOPBACK_FILE}`; else /sbin/losetup -f ${LOOPBACK_FILE}; fi"
+          ExecStop=bash -c "/sbin/losetup -d `/sbin/losetup -j ${LOOPBACK_FILE} -l -n -O NAME`"
+          RemainAfterExit=yes
+
+          [Install]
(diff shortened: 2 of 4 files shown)


config/samples/backends/lvm-nvme-tcp/backend.yaml
```diff
@@ -0,0 +1,27 @@
...
+            [lvm]
+            volume_backend_name=lvm_nvme_tcp
+            volume_driver=cinder.volume.drivers.lvm.LVMVolumeDriver
+            target_helper=nvmet
+            target_protocol=nvmet_tcp
+            target_port=4420
+            target_prefix=nvme-subsystem-1
+            nvmeof_conn_info_version=2
+            image_volume_cache_enabled=false
+            use_multipath_for_image_xfer=false
```
ASBishop, line 26: Why false?

config/samples/backends/lvm-nvme-tcp/backend.yaml
```diff
@@ -0,0 +1,27 @@
...
+          customServiceConfig: |
+            [lvm]
+            volume_backend_name=lvm_nvme_tcp
+            volume_driver=cinder.volume.drivers.lvm.LVMVolumeDriver
+            target_helper=nvmet
+            target_protocol=nvmet_tcp
+            target_port=4420
+            target_prefix=nvme-subsystem-1
+            nvmeof_conn_info_version=2
+            image_volume_cache_enabled=false
```
ASBishop, line 25: I'd like to know if we should disable the image volume cache in every backend sample. My concern is users may infer this is deliberate and not something they would be comfortable overriding. In other words, are the sample suggesting they shouldn't enable the cache?

config/samples/backends/lvm-nvme-tcp/backend.yaml
Akrog: Because NVMe-oF multipathing (ANA) is different from the multipathd approach, where I know that a single path works just fine. So until we have an isolated networks example with 2 networks and configuring the driver to do multipathing I'd rather not enable it.

config/samples/backends/lvm-nvme-tcp/backend.yaml
Akrog: For testing I always prefer to disable the cache because then during cleanup I don't have to remember to delete the cache volumes in the internal tenant project.
And I know it's the default, but I like things explicit.
Since this is a generic example for other people, and not just for me, I'll change it.

config/samples/backends/lvm-nvme-tcp/backend.yaml
ASBishop: Sure, that makes sense. It means we need to walk a fine line between these sample files being intended to document things for the user, versus the current state of our own development.

Is this an instance where a comment would help? Maybe something that states, "Disable multipath until it can be tested," or "TODO: enable multipath after it can be tested" (and you know I'm not a fan of seeing too many TODO in the code!)

config/samples/backends/lvm-nvme-tcp/lvm.yaml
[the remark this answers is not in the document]
ASBishop: Any thoughts on this?

config/samples/backends/lvm-nvme-tcp/backend.yaml
Akrog: Maybe we just need to document that these samples should not be taken as deployment "recommendations", just as working backend configurations and that they should look into the vendor's documentation for recommendations?

config/samples/backends/lvm-nvme-tcp/backend.yaml
Akrog: I can add a comment with something like:
```
# LVM NVMe-oF multipathing (ANA) requires special configuration to have multiple networks
# accessible from compute nodes and other cinder controller pods.
```

config/samples/backends/lvm-nvme-tcp/lvm.yaml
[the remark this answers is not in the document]
Akrog: Yes, I was thinking about doing a refactoring at some point, but I didn't want to make things confusing when looking at the directory structure, so I have to give it some more thought.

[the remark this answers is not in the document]
Akrog: This is the refactoring I have in mind: https://github.com/Akrog/cinder-operator/commit/66cc5ecf6be2d199a4f9cf06c4dfeafbd3e45939

config/samples/backends/lvm-nvme-tcp/backend.yaml
ASBishop: I'm OK leaving this as-is for now, and I think we can do a global tweak/cleanup when we get closer to GA.

config/samples/backends/lvm-nvme-tcp/lvm.yaml
```diff
@@ -0,0 +1,35 @@
...
+          [Service]
+          Environment=VG_GB_SIZE=10G
+          Environment=LOOPBACK_FILE=/var/home/core/cinder-volumes
+          Type=oneshot
+          ExecStart=bash -c "if [[ ! -e ${LOOPBACK_FILE} ]]; then /bin/truncate -s ${VG_GB_SIZE} ${LOOPBACK_FILE} && /sbin/vgcreate cinder-volumes `/sbin/losetup --show -f ${LOOPBACK_FILE}`; else /sbin/losetup -f ${LOOPBACK_FILE}; fi"
+          ExecStop=bash -c "/sbin/losetup -d `/sbin/losetup -j ${LOOPBACK_FILE} -l -n -O NAME`"
+          RemainAfterExit=yes
+
+          [Install]
+          WantedBy=local-fs-pre.target
```
ASBishop, line 35: The new linter they will introduce in pr 147 will complain about the lack of newline at end of file.

config/samples/backends/lvm-nvme-tcp/backend.yaml
Akrog: OK

config/samples/backends/lvm-nvme-tcp/lvm.yaml
Akrog: Fixed

config/samples/backends/lvm-nvme-tcp/nvme-fabrics.yaml
```diff
@@ -0,0 +1,26 @@
...
+          # Mode must be decimal, this is 0644
+          mode: 420
+          user:
+            name: root
+          group:
+            name: root
+          contents:
+            # Source can be a http, https, tftp, s3, gs, or data as defined in rfc2397.
+            # This is the rfc2397 text/plain string format
+            source: data:,nvme-fabrics
```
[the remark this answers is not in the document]
Akrog: Done

Akrog:
I will add the missing newlines to all the sample YAMLS in one of the patches I'll submit in the refactoring PR: https://github.com/Akrog/cinder-operator/commit/bc6de408318e74bd800260d0b6f9a4d83caf3b9f

ASBishop (approved):
/lgtm

PR #182, opened 2023-04-03
Remove leftover api sub module

Akrog:
This patch removes the leftover replacement of the lib-common library in the api sub module.


stuggi (approved):
/lgtm

PR #184, opened 2023-04-05
Refactor backend samples

Akrog:
Refactor our backend samples

This patch refactores how we structure our backend samples.

The idea is to have everything that can be common to multiple backends
in the `bases` directory, such as:

- Base `OpenStackControlPlane`
- iSCSI transport protocol requirement
- Multipathing transport protocol requirements
- NVMe-oF transport protocol requirements

Then we'll  have a specific directory for each Cinder driver.  At the
time of this writing these are:

- Ceph
- NFS
- LVM
  * Using iSCSI
  * Using NVMe-TCP

This PR also fixes missing newlines in our sample YAMLs.


config/samples/backends/bases/iscsid/iscsid.yaml
```diff
@@ -12,4 +12,5 @@ spec:
     systemd:
       units:
       - enabled: true
-        name: iscsid.service
\ No newline at end of file
+        name: iscsid.service
+
```
ASBishop, line 16: I think you added 2 newlines instead of just 1. L15 needs a newline, but L16 should be deleted.

config/samples/backends/bases/iscsid/kustomization.yaml
```diff
@@ -1,2 +1,3 @@
 resources:
   - iscsid.yaml
+
```
ASBishop, line 3: L2 was fine, so no additional newline required. L3 should be deleted.

config/samples/backends/bases/multipathd/multipathd.yaml
```diff
@@ -27,4 +27,5 @@ spec:
     systemd:
       units:
       - enabled: true
-        name: multipathd.service
\ No newline at end of file
+        name: multipathd.service
+
```
ASBishop, line 31: This is another instance where 1 newline was needed, not 2.

config/samples/backends/bases/lvm/kustomization.yaml
```diff
@@ -1,2 +1,3 @@
 resources:
   - lvm.yaml
+
```
ASBishop, line 3: ditto

config/samples/backends/bases/multipathd/kustomization.yaml
```diff
@@ -1,2 +1,3 @@
 resources:
   - multipathd.yaml
+
```
ASBishop, line 3: ditto

ASBishop (changes requested):
I think you added too many newlines. The only files that need them are these, and they just need a single newline in order to terminate the last line in the file.

config/samples/backends/bases/multipathd/multipathd.yaml
config/samples/backends/bases/iscsid/iscsid.yaml

ASBishop (approved):
/lgtm

PR #185, opened 2023-04-05
Remove fmt dependency from make tidy

gibizer:
The fmt dependency cannot be run if the go.mod file needs an update:
```
 $ make tidy
 go fmt ./...
 go: updates to go.mod needed; to update it:
 	go mod tidy
 make: *** [Makefile:103: fmt] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,7 +128,7 @@ vet: gowork ## Run go vet against code.
 	go vet ./api/...
 
 .PHONY: tidy
-tidy: fmt
+tidy: ## Run go mod tidy on every mod file in the repo
 	go mod tidy
 	cd ./api && go mod tidy
 


Makefile
```diff
@@ -128,7 +128,7 @@ vet: gowork ## Run go vet against code.
 	go vet ./api/...
 
 .PHONY: tidy
-tidy: fmt
+tidy: ## Run go mod tidy on every mod file in the repo
 	go mod tidy
```
Akrog, line 132: I believe the latest pre-commmit (`.pre-commit-config.yaml`) intended for us to run the `fmt` target.

So we probably need to update it and change line `args: ["tidy"]` to something `args: ["tidy", "fmt"]` right?

Akrog (changes requested):

Makefile
ASBishop: yeah, that makes sense

gibizer: As far as I see pre-commit runs golangci-lint https://github.com/openstack-k8s-operators/cinder-operator/blob/082c9624f769a83dba2bdef9a3ab4e53bf7f776f/.pre-commit-config.yaml#L1-L6 
And that is configured to run fmt: https://github.com/openstack-k8s-operators/cinder-operator/blob/082c9624f769a83dba2bdef9a3ab4e53bf7f776f/.golangci.yaml#L7

ASBishop (approved):
/lgtm

PR #186, opened 2023-04-06
Trim Cinder CRDs and optimize extraMounts usage

fmount:
The goal of this patch is to follow up on the Glance work [1] and give an alternative approach to what was proposed for Glance in [PR#184](https://github.com/openstack-k8s-operators/glance-operator/pull/184), where we tried to trim the `CRDs` as much as possible by injecting `extraMounts` from the top-level CR and get notified on a change coming from upper layer. However, the approach described in the Glance [PR#184](https://github.com/openstack-k8s-operators/glance-operator/pull/184) "breaks" the top-down pattern where parameters are specified in the `Spec` and passed to the underlying deployment, and in a way it "hides" the data from the requests. The current patch represents an implementation to solve the same problem using a different strategy: it trims less than what's described in `Glance` [PR#184](https://github.com/openstack-k8s-operators/glance-operator/pull/184) but keeps the pattern unchanged. In particular the `Cinder<Service>Spec` is built within the `Cinder` controller, but it's not explicitly included in the top level `Cinder` `API`. A `CinderServiceTemplate`, instead, is included and used to define the top-level input parameters provided and set to the sub-level resources. By doing this we can still keep the `ExtraMounts` definition in any `Cinder<Service>Spec`, but we don't replicate the same parameter in the top-level `CRD` (because the `Spec(s)` are not explicitly included).

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/191


controllers/cinder_controller.go
```diff
@@ -845,6 +845,24 @@ func (r *CinderReconciler) transportURLCreateOrUpdate(
 }
 
 func (r *CinderReconciler) apiDeploymentCreateOrUpdate(ctx context.Context, instance *cinderv1beta1.Cinder) (*cinderv1beta1.CinderAPI, controllerutil.OperationResult, error) {
+
+	apiTemplate := instance.Spec.CinderAPI
```
ASBishop, line 849: Two suggestions to consider:
1. Use the same API for creating all 4 of cinder's deployments, by passing in the service template as you do with volumeDeploymentCreateOrUpdate(). I know that the volume service requires you to pass in the service template because there will be separate calls for each backend. What you could do here is pass in the API service template (on L847) in lieu of the variable on L849.
2. I don't think it's necessary to embed the service's name in the variable name (or the volTemplate parameter name on L974). You could use something like "svcTemplate" so that the code blocks are similar in all 4 locations.

controllers/cinder_controller.go
```diff
@@ -845,6 +845,24 @@ func (r *CinderReconciler) transportURLCreateOrUpdate(
...
+	cinderAPISpec := cinderv1beta1.CinderAPISpec{
+		ContainerImage:         apiTemplate.ContainerImage,
+		Replicas:               apiTemplate.Replicas,
+		Debug:                  apiTemplate.Debug,
+		CustomServiceConfig:    apiTemplate.CustomServiceConfig,
```
ASBishop, line 854: I think you missed the CustomServiceConfigSecrets field (here, and in the backup, scheduler and volume functions as well).

controllers/cinder_controller.go
```diff
@@ -845,6 +845,24 @@ func (r *CinderReconciler) transportURLCreateOrUpdate(
...
+		DefaultConfigOverwrite: apiTemplate.DefaultConfigOverwrite,
+		NetworkAttachments:     apiTemplate.NetworkAttachments,
+		ServiceUser:            instance.Spec.ServiceUser,
+		DatabaseHostname:       instance.Status.DatabaseHostname,
+		DatabaseUser:           instance.Spec.DatabaseUser,
+		Secret:                 instance.Spec.Secret,
+		PasswordSelectors:      instance.Spec.PasswordSelectors,
+		ExtraMounts:            instance.Spec.ExtraMounts,
```
ASBishop, line 862: This confuses me. Won't you end up overwriting the service's own list of ExtraMounts with the top level CR's ExtraMounts?

controllers/cinder_controller.go
```diff
@@ -911,6 +931,22 @@ func (r *CinderReconciler) schedulerDeploymentCreateOrUpdate(ctx context.Context
 }
 
 func (r *CinderReconciler) backupDeploymentCreateOrUpdate(ctx context.Context, instance *cinderv1beta1.Cinder) (*cinderv1beta1.CinderBackup, controllerutil.OperationResult, error) {
+
+	bckTemplate := instance.Spec.CinderBackup
```
ASBishop, line 935: At first the "bck" prefix confused me, mainly because cinder tends to use "bak" to abbreviate the backup service. I'd request you switch to "bak" except, as I stated in an earlier comment, I think "svcTemplate" could be used in all 4 of the xxxDeploymentCreateOrUpdate functions.

controllers/cinder_controller.go
fmount: Good catch and sorry, I missed that! Adding the missing `CustomServiceConfigSecrets` in the next PS!

controllers/cinder_controller.go
fmount: +1 on replacing bck w/ bak to stay consistent.

controllers/cinder_controller.go
fmount: Thanks Alan for this comment: let me try to clarify this part (we can follow up offline in case we want to dig more).
For the single service (e.g. `Cinder{Api, Sched, Bak, Vol}`) we don't have to specify/override `ExtraVolumes`: we
 can always define them in the (`Cinder`) top-level `CR` and _"Propagate"_ them to the single service (it's what
 we called **"internal"** propagation). 
 Here's an example [1], where we can define two `extraVol`, one is destined to `Cinder`, which means it will be
 propagated to all the `Cinder` related pods/services, the other one is only destined to `volume1` instance and all
 the `CinderBackup` related pods.
 
 
**tl;dr** given the described approach, we only define `extraMounts` in the top-level `Cinder` `CR`:  each service
 inherits the struct from the "upper  layer" and filter the volumes according to the `Propagation` policy. The code
 change here doesn't change at all the previous behavior (look at [2], which is the same for API/Sched/Back/Vol).

[1] https://paste.opendev.org/show/bLQMPvZL3IfSzYFFN2Hh/
[2] https://github.com/fmount/cinder-operator/blob/master/controllers/cinder_controller.go#L864

controllers/cinder_controller.go
fmount: > Two suggestions to consider:
> 
>     1. Use the same API for creating all 4 of cinder's deployments, by passing in the service template as you do with volumeDeploymentCreateOrUpdate(). I know that the volume service requires you to pass in the service template because there will be separate calls for each backend. What you could do here is pass in the API service template (on L847) in lieu of the variable on L849.
> 
>     2. I don't think it's necessary to embed the service's name in the variable name (or the volTemplate parameter name on L974). You could use something like "svcTemplate" so that the code blocks are similar in all 4 locations.

The `CinderServiceTemplate` represents a common struct containing a list of fields that can be customized
 by the user for a given resource (`API/Sched/Bak/Vol`), because those can be changed/customized in the
 CR for the specific service: for instance, when the `CR` is defined, I can pass a diff `Resources` or `Replicas`
value for each Cinder service (or any other field provided by the `CinderServiceTemplate`).
This is the reason why I resolve the variable within the function (we pass `instance` here, and the templates
are already an attribute of its `Spec`: passing an additional variable containing data that we already pass
seems redundant to me).
Also, I don't always use `API` in all the 4 functions: each of them get the template values with the form:

```
<service>Template := instance.Spec.<Service>
```

For the reason above I'd like to keep this variable within the function, unless we have strong reasons to resolve
it outside and pass it to the function.

Regarding \#2, volumes are defined in the form:
```
CinderVolumes map[string]CinderServiceTemplate
```

which means that the `name` we extract in the `for loop` is the `key` of this map, and it represents the instance
name we want to pass to the k8s `Deployment`: passing it to the `VolCreateOrUpdate` function looks good to
me as we want to process,  for each backend, the associated parameters, and return the `CinderVolume` instance):
it's used, later in the code (but within the same for loop) to compute the status (`ReadyCount`, etc ..) and change the `Cinder.Status` accordingly.
I think it would result confusing changing the current logic, embedding the name (key) within a struct which represents
the associated value.

controllers/cinder_controller.go
fmount: Done, feel free to Resolve the conversation if the change it's ok.

controllers/cinder_controller.go
fmount: Done, feel free to resolve the conversation if this part it's ok.

abays:
> @fmount: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> Test name 	Commit 	Details 	Required 	Rerun command
> ci/prow/cinder-operator-build-deploy-kuttl 	[8d2d477](https://github.com/openstack-k8s-operators/cinder-operator/commit/8d2d477948cf56a9d368a559e5432bd5f221370d) 	[link](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_cinder-operator/186/pull-ci-openstack-k8s-operators-cinder-operator-master-cinder-operator-build-deploy-kuttl/1644280755691458560) 	false 	`/test cinder-operator-build-deploy-kuttl`
> 
> [Full PR test history](https://prow.ci.openshift.org/pr-history?org=openstack-k8s-operators&repo=cinder-operator&pr=186). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:fmount).

It looks like `Cinder` never reaches the `Ready` state, and then CI eventually times-out:

(From `OpenStackControlPlane` CR `Status`)
```
    - lastTransitionTime: "2023-04-07T10:54:44Z"
      message: OpenStackControlPlane Cinder in progress
      reason: Requested
      severity: Info
      status: "False"
      type: OpenStackControlPlaneCinderReady
```

(From CI logs)
```
+ timeout 1200 xargs -d '\n' -I '{}' sh -c 'echo testing condition={}; oc wait openstackcontrolplane.core.openstack.org/openstack --for=condition={} --timeout=-1s'
testing condition=Ready
```

abays:
> > @fmount: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> > Test name 	Commit 	Details 	Required 	Rerun command
> > ci/prow/cinder-operator-build-deploy-kuttl 	[8d2d477](https://github.com/openstack-k8s-operators/cinder-operator/commit/8d2d477948cf56a9d368a559e5432bd5f221370d) 	[link](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_cinder-operator/186/pull-ci-openstack-k8s-operators-cinder-operator-master-cinder-operator-build-deploy-kuttl/1644280755691458560) 	false 	`/test cinder-operator-build-deploy-kuttl`
> > [Full PR test history](https://prow.ci.openshift.org/pr-history?org=openstack-k8s-operators&repo=cinder-operator&pr=186). [Your PR dashboard](https://prow.ci.openshift.org/pr?query=is:pr+state:open+author:fmount).
> 
> It looks like `Cinder` never reaches the `Ready` state, and then CI eventually times-out:
> 
> (From `OpenStackControlPlane` CR `Status`)
> 
> ```
>     - lastTransitionTime: "2023-04-07T10:54:44Z"
>       message: OpenStackControlPlane Cinder in progress
>       reason: Requested
>       severity: Info
>       status: "False"
>       type: OpenStackControlPlaneCinderReady
> ```
> 
> (From CI logs)
> 
> ```
> + timeout 1200 xargs -d '\n' -I '{}' sh -c 'echo testing condition={}; oc wait openstackcontrolplane.core.openstack.org/openstack --for=condition={} --timeout=-1s'
> testing condition=Ready
> ```

It looks like the `StatefulState` for the `CinderBackup` never reaches the `Ready` state:

(From the `cinder-backup` `StatefulSet.Status`)
```
status:
  availableReplicas: 0
  collisionCount: 0
  currentRevision: cinder-backup-76747b8d8f
  observedGeneration: 1
  replicas: 0
  updateRevision: cinder-backup-76747b8d8f
```

I'm not sure where in the existing `must-gather` logs we need to look to determine what exactly is preventing the `StatefulSet` from spawning its pod.

controllers/cinder_controller.go
ASBishop: Thanks, this explanation really helped. I had forgotten the whole ExtraVolumes concept includes its own mechanism for propagating volumes to specific CRs. This is why you define your ExtraVolumes at the top, and copy them into the lower layers.

controllers/cinder_controller.go
ASBishop: I think we're in complete agreement on what the code is doing (I agree with everything you wrote), and the issue is one of coding style. I typically defer to the person who wrote the code, so I'm OK with the code as is (though I'd like Gorka to see this discussion).

BTW, the reason why I suggested using "svcConfig" for the local variable name (and function parameter in the Volume code) is it helps highlight common blocks of code. And the reason why I spotted this is that Gorka and I hope to (eventually) dedup code wherever possible.

controllers/cinder_controller.go
```diff
@@ -911,6 +933,23 @@ func (r *CinderReconciler) schedulerDeploymentCreateOrUpdate(ctx context.Context
...
+		CustomServiceConfigSecrets: bakTemplate.CustomServiceConfigSecrets,
+		DefaultConfigOverwrite:     bakTemplate.DefaultConfigOverwrite,
+		NetworkAttachments:         bakTemplate.NetworkAttachments,
+		ServiceUser:                instance.Spec.ServiceUser,
+		DatabaseHostname:           instance.Status.DatabaseHostname,
+		DatabaseUser:               instance.Spec.DatabaseUser,
+		Secret:                     instance.Spec.Secret,
+		ExtraMounts:                instance.Spec.ExtraMounts,
+		TransportURLSecret:         instance.Status.TransportURLSecret,
+	}
```
ASBishop, line 952: kuttl deployment failed because it's not propagating c-bak's PasswordSelectors

ASBishop (changes requested):
You can see c-bak's PasswordSelectors are empty/blank in [1]

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_cinder-operator/186/pull-ci-openstack-k8s-operators-cinder-operator-master-cinder-operator-build-deploy-kuttl/1644280755691458560/artifacts/cinder-operator-build-deploy-kuttl/openstack-k8s-operators-gather/artifacts/api-resources/cinderbackups.yaml

controllers/cinder_controller.go
fmount: Thanks, Good catch, I have missed that in bak, and this probably highlight the need of avoiding code duplication whenever is possible (it would avoid this kind of issues).

controllers/cinder_controller.go
fmount: > I think we're in complete agreement on what the code is doing (I agree with everything you wrote), and the issue is one of coding style. I typically defer to the person who wrote the code, so I'm OK with the code as is (though I'd like Gorka to see this discussion).

+1

> 
> BTW, the reason why I suggested using "svcConfig" for the local variable name (and function parameter in the Volume code) is it helps highlight common blocks of code. And the reason why I spotted this is that Gorka and I hope to (eventually) dedup code wherever possible.

I'm ok to avoid duplication as much as possible (it would avoid the issues of the previous PS), and I think we have a couple of options to do that (just not sure we might want to have a follow up change on that part).
1. Try using generics: provide a function that accepts an interface and build the `Cinder<Service>Spec`; this is feasible but I think it would result hard to read/maintain, other that the fact we might still need logic to distinguish in between the Service passed/returned for a few specific fields
2. Embed the common fields of the `Cinder<Service>Spec` in the `CinderServiceTemplate` struct which can be defined in the services Spec itself.

E.g.

 ```
type CinderAPISpec {
   ...
   ...
   // +kubebuilder:validation:Required
   // CinderServiceTemplate specifies the generic fields of the service
   CinderServiceTemplate `json:",inline"`
  ...
```

and in the controller we can have something like:

```
	cinderAPISpec := cinderv1beta1.CinderAPISpec{
		CinderServiceTemplate: cinderv1beta1.CinderServiceTemplate{
			ContainerImage:             svcTemplate.ContainerImage,
			Replicas:                   svcTemplate.Replicas,
			Debug:                      svcTemplate.Debug,
			CustomServiceConfig:        svcTemplate.CustomServiceConfig,
			DefaultConfigOverwrite:     svcTemplate.DefaultConfigOverwrite,
			CustomServiceConfigSecrets: svcTemplate.CustomServiceConfigSecrets,
			NetworkAttachments:         svcTemplate.NetworkAttachments,
		},
		ServiceUser:                instance.Spec.ServiceUser,
		DatabaseHostname:           instance.Status.DatabaseHostname,
		DatabaseUser:               instance.Spec.DatabaseUser,
		Secret:                     instance.Spec.Secret,
		PasswordSelectors:          instance.Spec.PasswordSelectors,
		ExtraMounts:                instance.Spec.ExtraMounts,
		TransportURLSecret:         instance.Status.TransportURLSecret,
	}
```

This means, in theory, that the struct we pass in the above can be computed by a generic function, which is valid for all the Cinder services: this would avoid a lot of duplication given that the code would become:

```
svcTemplate := r.getServiceTemplate(instance.Spec.CinderAPI)
cinderAPISpec := cinderv1beta1.CinderAPISpec{
		CinderServiceTemplate: svcTemplate,
		ServiceUser:                instance.Spec.ServiceUser,
		DatabaseHostname:           instance.Status.DatabaseHostname,
		DatabaseUser:               instance.Spec.DatabaseUser,
		Secret:                     instance.Spec.Secret,
		PasswordSelectors:          instance.Spec.PasswordSelectors,
		ExtraMounts:                instance.Spec.ExtraMounts,
		TransportURLSecret:         instance.Status.TransportURLSecret,
	}
```

and that function should be something like:

```
func (r *CinderReconciler) getServiceTemplate(instance cinderv1beta1.CinderSpec) cinderv1beta1.CinderServiceTemplate {
	switch req {
	case "API":
		return instance.CinderAPI
	case "Scheduler":
		return instance.CinderScheduler
        ...
        ...
```

However, I'd like to keep a change like this for a follow up patch, because it doesn't change much on the CRD side (which is the purpose of this patch) and can be reviewed easily because it would be smaller and focused on the optimization front.

Comment (ASBishop) [reply controllers/cinder_controller.go]: I'm not concerned about deduping in this patch, so I'm fine with this patch as is. I think you can leave the dedup concerns to Gorka and me.

File: api/v1beta1/common_types.go [hunk 6]
```diff
@@ -16,7 +16,65 @@ limitations under the License.
...
+	// https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// NetworkAttachments is a list of NetworkAttachment resource names to expose the services to the given network
+	NetworkAttachments []string `json:"networkAttachments,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// ExternalEndpoints, expose a VIP via MetalLB on the pre-created address pool
+	ExternalEndpoints []MetalLBConfig `json:"externalEndpoints,omitempty"`
```
gibizer, line 76: // I know this is already merged so this is for a follow up
Do we really need ExternalEndpoints for non API services in cinder?

controllers/cinder_controller.go
```diff
@@ -845,6 +845,25 @@ func (r *CinderReconciler) transportURLCreateOrUpdate(
...
+		DefaultConfigOverwrite:     svcTemplate.DefaultConfigOverwrite,
+		CustomServiceConfigSecrets: svcTemplate.CustomServiceConfigSecrets,
+		NetworkAttachments:         svcTemplate.NetworkAttachments,
+		ServiceUser:                instance.Spec.ServiceUser,
+		DatabaseHostname:           instance.Status.DatabaseHostname,
+		DatabaseUser:               instance.Spec.DatabaseUser,
+		Secret:                     instance.Spec.Secret,
+		PasswordSelectors:          instance.Spec.PasswordSelectors,
+		ExtraMounts:                instance.Spec.ExtraMounts,
+		TransportURLSecret:         instance.Status.TransportURLSecret,
```
gibizer, line 864: I think you missed ExternalEndpoints

controllers/cinder_controller.go
```diff
@@ -878,6 +889,25 @@ func (r *CinderReconciler) apiDeploymentCreateOrUpdate(ctx context.Context, inst
...
+		CustomServiceConfigSecrets: svcTemplate.CustomServiceConfigSecrets,
+		DefaultConfigOverwrite:     svcTemplate.DefaultConfigOverwrite,
+		NetworkAttachments:         svcTemplate.NetworkAttachments,
+		ServiceUser:                instance.Spec.ServiceUser,
+		DatabaseHostname:           instance.Status.DatabaseHostname,
+		DatabaseUser:               instance.Spec.DatabaseUser,
+		Secret:                     instance.Spec.Secret,
+		ExtraMounts:                instance.Spec.ExtraMounts,
+		TransportURLSecret:         instance.Status.TransportURLSecret,
+		PasswordSelectors:          instance.Spec.PasswordSelectors,
```
gibizer, line 908: We need to ExternalEndpoints here as well or we need to remove it from the common template if that is not a common field

controllers/cinder_controller.go
```diff
@@ -911,6 +933,24 @@ func (r *CinderReconciler) schedulerDeploymentCreateOrUpdate(ctx context.Context
...
+		CustomServiceConfigSecrets: svcTemplate.CustomServiceConfigSecrets,
+		DefaultConfigOverwrite:     svcTemplate.DefaultConfigOverwrite,
+		NetworkAttachments:         svcTemplate.NetworkAttachments,
+		ServiceUser:                instance.Spec.ServiceUser,
+		DatabaseHostname:           instance.Status.DatabaseHostname,
+		DatabaseUser:               instance.Spec.DatabaseUser,
+		PasswordSelectors:          instance.Spec.PasswordSelectors,
+		Secret:                     instance.Spec.Secret,
+		ExtraMounts:                instance.Spec.ExtraMounts,
+		TransportURLSecret:         instance.Status.TransportURLSecret,
```
gibizer, line 952: same as above about the externalEndpoints

controllers/cinder_controller.go
```diff
@@ -943,7 +975,25 @@ func (r *CinderReconciler) backupDeploymentCreateOrUpdate(ctx context.Context, i
...
+		CustomServiceConfigSecrets: svcTemplate.CustomServiceConfigSecrets,
+		DefaultConfigOverwrite:     svcTemplate.DefaultConfigOverwrite,
+		NetworkAttachments:         svcTemplate.NetworkAttachments,
+		Secret:                     instance.Spec.Secret,
+		ExtraMounts:                instance.Spec.ExtraMounts,
+		DatabaseHostname:           instance.Status.DatabaseHostname,
+		DatabaseUser:               instance.Spec.DatabaseUser,
+		ServiceUser:                instance.Spec.ServiceUser,
+		TransportURLSecret:         instance.Status.TransportURLSecret,
+		PasswordSelectors:          instance.Spec.PasswordSelectors,
```
gibizer, line 994: ExternalEndpoints

controllers/cinder_controller.go
fmount: (see comment above)

controllers/cinder_controller.go
fmount: (see comment above)

controllers/cinder_controller.go
fmount: (see comment above)

api/v1beta1/common_types.go
gibizer: Thanks. the follow up looks good.

PR #187, opened 2023-04-06
Fix cinder-api pod name

ASBishop:
Use instance.Name so that "-api" will appear in the pod name. Now c-api pods will be named like this:

  cinder-api-58ccfdc568-6lg6m

whereas they previously looked like this:

  cinder-58ccfdc568-6lg6m


abays:
If we do this, we'll have to update the KUTTL tests as well.  From the CI logs:

```
case.go:364: failed in step 1-deploy-cinder
case.go:366: deployments.apps "cinder" not found
```

They're still looking for the old `cinder.ServiceName`.

abays (approved):
/lgtm

PR #190, opened 2023-04-11
Use composition to dedup structure definitions

ASBishop:
Define common structures to reduce the need to repeat a lot of fields in the top level Cinder and Cinder service CRDs. This will also help future efforts to dedup code that accesses the structures.

Where necessary, services can add additional fields to its service specific CRD. For example:
- Replicas are service specific, and CinderVolume Replicas are constrained to Maximum=1 (until we resolve how to support drivers that can run active/active).
- The CinderAPI CRD supports ExternalEndpoints.


api/v1beta1/cinder_types.go
```diff
@@ -32,38 +32,20 @@ const (
 
 // CinderSpec defines the desired state of Cinder
 type CinderSpec struct {
-	// +kubebuilder:validation:Optional
-	// +kubebuilder:default=cinder
-	// ServiceUser - optional username used for this service to register in cinder
-	ServiceUser string `json:"serviceUser"`
+	CinderTemplate `json:",inline"`
```
fmount, line 35: +1 on composing the Spec, which is something I had in mind with the usage of `json:",inline"`.

api/v1beta1/cinderapi_types.go
```diff
@@ -18,93 +18,46 @@ package v1beta1
...
-	// +kubebuilder:validation:Optional
-	// +kubebuilder:default=cinder
-	// ServiceUser - optional username used for this service to register in cinder
-	ServiceUser string `json:"serviceUser"`
-
-	// +kubebuilder:validation:Required
-	// ContainerImage - Cinder API Container Image URL (will be set to environmental default if empty)
-	ContainerImage string `json:"containerImage"`
+// CinderAPITemplate defines the input parameters for the Cinder API service
+type CinderAPITemplate struct {
```
fmount, line 25: Adding this "middle layer" here allows to resolve the struct directly in the controller (`instance.Spec.CinderAPI` will include the `CinderServiceTemplate`, which collect the user input in the service CR).

controllers/cinder_controller.go
```diff
@@ -846,22 +846,12 @@ func (r *CinderReconciler) transportURLCreateOrUpdate(
...
-		NetworkAttachments:         svcTemplate.NetworkAttachments,
-		ServiceUser:                instance.Spec.ServiceUser,
-		DatabaseHostname:           instance.Status.DatabaseHostname,
-		DatabaseUser:               instance.Spec.DatabaseUser,
-		Secret:                     instance.Spec.Secret,
-		PasswordSelectors:          instance.Spec.PasswordSelectors,
-		ExtraMounts:                instance.Spec.ExtraMounts,
-		TransportURLSecret:         instance.Status.TransportURLSecret,
+		CinderTemplate:     instance.Spec.CinderTemplate,
+		CinderAPITemplate:  instance.Spec.CinderAPI,
```
fmount, line 851: +1 here, using `CinderAPITemplate` we can easily shorten the code here (same comment applies to the other services).

controllers/cinder_controller.go
```diff
@@ -890,22 +880,12 @@ func (r *CinderReconciler) apiDeploymentCreateOrUpdate(ctx context.Context, inst
...
-		DatabaseHostname:           instance.Status.DatabaseHostname,
-		DatabaseUser:               instance.Spec.DatabaseUser,
-		Secret:                     instance.Spec.Secret,
-		ExtraMounts:                instance.Spec.ExtraMounts,
-		TransportURLSecret:         instance.Status.TransportURLSecret,
-		PasswordSelectors:          instance.Spec.PasswordSelectors,
+		CinderTemplate:          instance.Spec.CinderTemplate,
+		CinderSchedulerTemplate: instance.Spec.CinderScheduler,
+		ExtraMounts:             instance.Spec.ExtraMounts,
+		DatabaseHostname:        instance.Status.DatabaseHostname,
```
fmount, line 887: I was concerned about this field at the beginning, because it comes from the upper layer `Status`, but I see you still explicitly set it here.

fmount (approved):
Looks good from me, and it recalls something I had in mind in PR #186, where I was about to shorten this code using a similar approach.
Thanks Alan!

PR #191, opened 2023-04-11
Ctrl manager labels for webhooks and logging

abays:
For initial webhook work, we found that we needed to remove the `control-plane: controller-manager` label and label-selector from the `controller-manager` pod and its associated peripherals in order to properly direct webhook traffic to the appropriate webhook server for our various operators' CRDs.  This had the unfortunate side effect of removing a label that could be used for bulk-selecting all our `controller-manager` pods.  This was noted here [1].  We therefore have opted to reintroduce the `control-plane: controller-manager` default label (originally added by `operator-sdk` scaffolding) and instead use another unique label (also added by `operator-sdk` scaffolding, at least in newer version of `operator-sdk`) for routing webhook traffic.  This unique label takes the form of `app.kubernetes.io/name: <operator name>` and can be used with label-selectors to ensure webhooks and other peripheral k8s services work correctly.

[1] https://github.com/openshift/release/pull/37084#discussion_r1138314478

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -4,18 +4,22 @@ metadata:
   name: controller-manager
   namespace: system
   labels:
-    control-plane: cinder-controller-manager
+    control-plane: controller-manager
+    app.kubernetes.io/name: cinder-operator
+    app.kubernetes.io/component: cinder
 spec:
   selector:
     matchLabels:
-      control-plane: cinder-controller-manager
+      app.kubernetes.io/name: cinder-operator
   replicas: 1
   template:
     metadata:
       annotations:
         kubectl.kubernetes.io/default-container: manager
       labels:
-        control-plane: cinder-controller-manager
+        control-plane: controller-manager
+        app.kubernetes.io/name: cinder-operator
+        app.kubernetes.io/component: cinder
     spec:
       securityContext:
         runAsNonRoot: true
--- a/config/prometheus/monitor.yaml
+++ b/config/prometheus/monitor.yaml
@@ -4,7 +4,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   labels:
-    control-plane: cinder-controller-manager
+    control-plane: controller-manager
   name: controller-manager-metrics-monitor
   namespace: system
 spec:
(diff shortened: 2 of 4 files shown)


config/manager/manager.yaml
```diff
@@ -4,18 +4,22 @@ metadata:
   name: controller-manager
   namespace: system
   labels:
-    control-plane: cinder-controller-manager
+    control-plane: controller-manager
+    app.kubernetes.io/name: cinder-operator
+    app.kubernetes.io/component: cinder
```
ASBishop, line 9: Where is this 'app.kubernetes.io/component' label used?

While it may not be a big deal, we already use a generic 'component' label on several resources that designates the specific cinder service (API, Backup, Scheduler, Volume). I don't know if others will be confused when we use the term "component" in a different way that depends on the context. In the app.kubernetes context, "component" means cinder, but in the cinder context, "component" means [API|Backup|Scheduler|Volume].

I'm not saying this is a problem, but am curious how others thing about this.

abays: To be honest, the `app.kubernetes.io/component: <openstack service name>` label doesn't seem to be used anywhere within the operator itself.  It would be the same case for the `app.kubernetes.io/name: <openstack service name>-operator` label if we were not now using it as a unique label selector.  The `operator-sdk` added these sorts of labels to other operators (Keystone and Placement, for example) and for some reason they were missing here and in other operators.  My guess is that these labels are added via newer versions of `operator-sdk`.  Since this Cinder operator was actually scaffolded back in 2020, the labels are simply missing.  I suppose we could exclude adding `app.kubernetes.io/component: cinder` since it's not referenced elsewhere though.  It just initially seemed prudent to repeat the newer scaffolded pattern and include it.  I wonder if these new labels are added by the scaffolding because newer versions of OLM will try to look for them and thus use them for certain operations?

All that being said, I see what you mean regarding the potential "component" confusion (as an example: [1]).  I'm not immediately sure what the right approach is, whether that's leaving it as-is, explicitly excluding/removing it, adding a comment to help dispel ambiguity, etc.

[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/6ebfd7a14c4b315b48705b8d69a6b076671ebcdb/controllers/cinderapi_controller.go#L554

fmount (approved):

PR #196, opened 2023-04-21
Add cinder-backup mount for /var/lib/cinder

ASBishop:
A previous patch created the appropriate volume so that cinder-backup has access to /var/lib/cinder, but it neglected to mount the volume.


Akrog (approved):
/lgtm

PR #198, opened 2023-04-25
Dynamic serviceaccount/role/rolebinding creation

dprince:
This PR dyamically creates the above resources within the namespace of the requested Cinder instance.

The above resources are no longer created via OLM.

This allows the operator to deploy openstack service into any namespace within the OpenShift cluster.


abays (commented):
+1 from me

fmount (approved):
It's basically aligned to [1]  and the CI around this change is green

[1] https://github.com/openstack-k8s-operators/glance-operator/pull/210

api/v1beta1/cinderapi_types.go
```diff
@@ -58,6 +58,10 @@ type CinderAPISpec struct {
 	// +kubebuilder:validation:Optional
 	// ExtraMounts containing conf files and credentials
 	ExtraMounts []CinderExtraVolMounts `json:"extraMounts,omitempty"`
+
+	// +kubebuilder:validation:Required
+	// ServiceAccount - service account name used internally to provide Cinder services the default SA name
+	ServiceAccount string `json:"serviceAccount"`
```
ASBishop, line 64: Sorry that I'm late to the party (code merged already), but I wonder if these common fields should be included in the CinderServiceTemplate struct. We've been trying to dedup a lot of redundant data structures and code. @fmount I think you are more familiar with this effort, and am curious to hear your thoughts.

controllers/cinder_controller.go
```diff
@@ -852,6 +887,7 @@ func (r *CinderReconciler) apiDeploymentCreateOrUpdate(ctx context.Context, inst
 		ExtraMounts:        instance.Spec.ExtraMounts,
 		DatabaseHostname:   instance.Status.DatabaseHostname,
 		TransportURLSecret: instance.Status.TransportURLSecret,
+		ServiceAccount:     instance.RbacResourceName(),
```
ASBishop, line 890: Per my previous comment, if the ServiceAccount were part of the CinderServiceTemplate struct, then this field would be automatically populated by L886. Lines L890, L925, L960 and L994 could be eliminated.

ASBishop (commented):
+1, though I see an opportunity to dedup some stuff.

api/v1beta1/cinderapi_types.go
fmount: @ASBishop in theory you're right in terms of data structure it can easily be absorbed in  `CinderServiceTemplate`. However, given it's not a field populated by the `Spec` (but the result of `RbacResourceName()` function), we should properly populate this field when the Template is generated.
This was the reason `DatabaseHostname` and a few other fields are still part of the Spec instead of being passed from the upper layer, but I agree we can follow up providing a unified strategy that allows more dedup.

PR #205, opened 2023-05-17
Validate input for Replicas spec

kajinamit:
This adds minimum/maximum value for Replicas spec to avoid accepting a negative number or a too large number. The maximum number follows what is used in keystone-operator.

### DIFF
--- a/api/v1beta1/cinderapi_types.go
+++ b/api/v1beta1/cinderapi_types.go
@@ -28,6 +28,8 @@ type CinderAPITemplate struct {
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=1
+	// +kubebuilder:validation:Maximum=32
+	// +kubebuilder:validation:Minimum=0
 	// Replicas - Cinder API Replicas
 	Replicas int32 `json:"replicas"`
 
--- a/api/v1beta1/cinderscheduler_types.go
+++ b/api/v1beta1/cinderscheduler_types.go
@@ -28,6 +28,8 @@ type CinderSchedulerTemplate struct {
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=1
+	// +kubebuilder:validation:Maximum=1
+	// +kubebuilder:validation:Minimum=0
 	// Replicas - Cinder Scheduler Replicas
 	Replicas int32 `json:"replicas"`
 }
--- a/api/bases/cinder.openstack.org_cinderapis.yaml
+++ b/api/bases/cinder.openstack.org_cinderapis.yaml
@@ -871,6 +871,8 @@ spec:
               replicas:
                 default: 1
                 format: int32
+                maximum: 32
+                minimum: 0
                 type: integer
               resources:
                 properties:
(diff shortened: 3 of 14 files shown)


api/v1beta1/cinderapi_types.go
```diff
@@ -28,6 +28,8 @@ type CinderAPITemplate struct {
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=1
+	// +kubebuilder:validation:Maximum=32
```
ASBishop, line 31: Why so many? Tripleo generally deploys 3 (one per controller).

api/v1beta1/cinderscheduler_types.go
```diff
@@ -28,6 +28,8 @@ type CinderSchedulerTemplate struct {
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=1
+	// +kubebuilder:validation:Maximum=32
```
ASBishop, line 31: We will likely want to limit the deployment to a single scheduler. Tests have shown a single scheduler can handle the load, and having only one scheduler will avoid a class of quota bugs that arise when multiple schedulers are processing volume requests in parallel. I'd prefer to set the maximum to 1, and we can raise it later if a single scheduler can't handle the load.

api/v1beta1/cinderapi_types.go
kajinamit: This follows the value commonly used in the other operators.

Also, because now we do not we provide a way to configure wsgi processes/threads, it's more likely that users need to scale out api pods. So having a relatively high maximum makes sense, IMO.

api/v1beta1/cinderscheduler_types.go
kajinamit: One concern with enforcing maximum 1 is that we have to release a new operator to allow users to do scale out if they need it.

However probably c-vol would become bottleneck even if we scale scheduler pods and we unlikely need to scale only scheduler.

kajinamit: Updated the value accordingly.

Akrog (changes requested):
I'm ok with setting the minimum to 0, but until we do proper performance testing in Cinder we don't want to limit the maximum in any way.

kajinamit:
> I'm ok with setting the minimum to 0, but until we do proper performance testing in Cinder we don't want to limit the maximum in any way.

Is this about maximum 1 we discussed or do you disagree with max: 32 ?
I think 32 is quite large and very likely enough for any cases so I tend to leave that "consistent" upper limit.

Akrog:
> Is this about maximum 1 we discussed or do you disagree with max: 32 ? I think 32 is quite large and very likely enough for any cases so I tend to leave that "consistent" upper limit.

It's about setting a maximum at all, because it begs the question "why?" and to answer that question we would have to provide actual data, not just say "it seemed right at the moment".

So I'm against setting a maximum altogether.  Except for cinder-volume where setting it higher than 1 has other implications.

PR #210, opened 2023-05-31
Reset Ready condition when update failed

kajinamit:
This ensures the Ready Condition is reset when CR becomes once ready but something fails during reconfiguration.

### DIFF
--- a/api/v1beta1/cinderapi_types.go
+++ b/api/v1beta1/cinderapi_types.go
@@ -113,7 +113,7 @@ func init() {
 	SchemeBuilder.Register(&CinderAPI{}, &CinderAPIList{})
 }
 
-// IsReady - returns true if service is ready to serve requests
+// IsReady - returns true if CinderAPI is reconciled successfully
 func (instance CinderAPI) IsReady() bool {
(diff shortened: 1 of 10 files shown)


api/v1beta1/cinderapi_types.go
```diff
@@ -113,7 +113,7 @@ func init() {
 	SchemeBuilder.Register(&CinderAPI{}, &CinderAPIList{})
 }
 
-// IsReady - returns true if service is ready to serve requests
+// IsReady - returns true if CinderAPI is reconciled successfully
 func (instance CinderAPI) IsReady() bool {
-	return instance.Status.ReadyCount == instance.Spec.Replicas
```
kajinamit, line 118: So current IsReady requires that requested number of pods are ready, but (Deployment)ReadyCondition is set to true if at least one pod is available.

My understanding is these services should work as long as at least one pod is available but please correct me if there is any additional requirements.

Akrog (commented):
Thanks for the fix.
This PR is also changing the meaning of Ready (currently we need all replicas to be running), and while the cinder-operator may be incorrectly understanding the term, we would like to have a discussion on the weekly meeting before proceeding.

PR #212, opened 2023-06-09
Revamp config process

ASBishop:
This is an implementation for cinder that's consistent with the [Propose Service Bootstrap config design](https://github.com/openstack-k8s-operators/docs/pull/31#top)

This PR consists of 3 patches, the first of which contains the most technically relevant changes, followed by 2 more "cleanup" patches. When reviewing the first patch, if you see some "cruft" that should be cleaned up, bear in mind it might be handled in one of the subsequent patches.


Akrog (approved):
/lgtm

PR #217, opened 2023-06-23
Stop mouting localtime from host

kajinamit:
In general we should avoid mounting hostpath in k8s pods. Timezone can be configured using TZ environment and this is more flexible. (For example we can configure timezone different from the one used in the platform.)

### DIFF
--- a/pkg/cinder/volumes.go
+++ b/pkg/cinder/volumes.go
@@ -24,14 +24,6 @@ func GetVolumes(name string, storageSvc bool, extraVol []cinderv1beta1.CinderExt
 			},
 		},
 		{
-			Name: "etc-localtime",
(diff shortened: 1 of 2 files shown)


Akrog:
Why would we want to configure it differently than the host? I don't see real value on that.
And this also adds extra work to actually set the env variables, right?

kajinamit:
> Why would we want to configure it differently than the host? I don't see real value on that.
The main concern is that mounting this file (and instance-id) requires privilege mode enabled.
Using TZ env is the standard way to customized timezone inside containers as suggested in https://access.redhat.com/solutions/4994241 .

> And this also adds extra work to actually set the env variables, right?
It's true, though we probably leave actual implementation now unless we actually have a use case where users need to customize timezone used in each service log.

PR #218, opened 2023-06-23
Backup: Enable multipathing by default

Akrog:
The default for cinder volume service is to enable multipathing, but we are missing this on the backup service.

This patch enables multipathing by default for backup services as well.


ASBishop (approved):
/lgtm

PR #220, opened 2023-06-28
Refresh "openstack" backend base sample config

ASBishop:
This realigns the sample config with changes made to other service's CRD, namely OVS/OVN networking.


abays (approved):
/lgtm

PR #221, opened 2023-06-29
Fix 'latest' tag used in GitHub automated builds

abays:
When we switched from `master` to `main` as the default branch name, we forgot to update the GitHub workflow scripts for automated builds.  We've been mistakenly building `main-latest` instead of `latest` for the past two months.  This was at least causing `install_yamls` to use the outdated version since then.


fmount (approved):
Thanks!

PR #222, opened 2023-06-30
DBSync: Bump service versions

Akrog:
When adopting the cinder services we end up with the services getting pinned to the old RPC and OVO versions, and to fix that we would need to restart the services.  This is the standard rolling upgrades mechanism.

Since the adoption is more like an offline update/upgrade we need to bump these versions in the DB before we start them so they all start with the right versions.

The `cinder-manage db sync` command has a flag specifically designed for this purpose: `--bump-versions`

This patch adds this flag to the job that syncs the DB.

Once we start working on updates and upgrades we'll have to revisit this to see if we need to tweak it or not.

### DIFF
--- a/pkg/cinder/dbsync.go
+++ b/pkg/cinder/dbsync.go
@@ -11,6 +11,13 @@ import (
 
 const (
 	// DBSyncCommand -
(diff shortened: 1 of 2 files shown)


abays (approved):
/lgtm

PR #223, opened 2023-06-30
Fix cinder-api logging to stdout

Akrog:
On commit 64a5e288ef977be47fead7941d5489d72142f8f2 we made the logs go to /dev/stdout to make the API service logs not appear as [wsgi:error], but on some deployments we now find that the Cinder-API service returns status code 500 and doesn't work at all.

Inside the API logs we see error:
  PermissionError: [Errno 13] Permission denied: '/dev/stdout'

Recurrently looking at permissions of /dev/stdout we find:

  [cinder@cinder-api-54859ccfbd-pqxwf ~]$ ls -la /dev/stdout
  lrwxrwxrwx. 1 root root 15 Jun 28 19:41 /dev/stdout -> /proc/self/fd/1

  [cinder@cinder-api-54859ccfbd-pqxwf ~]$ ls -la /proc/self/fd/1
  lrwx------. 1 cinder cinder 64 Jun 28 20:01 /proc/self/fd/1 -> /dev/pts/0

  [cinder@cinder-api-54859ccfbd-pqxwf ~]$ ls -la /dev/pts/0
  crw--w----. 1 root tty 136, 0 Jun 28 20:01 /dev/pts/0

Other cinder-services (scheduler, volume, and backup) don't have this issue because they run as root, but httpd doesn't allow to run mod_wsgi as root, so we are running it as cinder, as seen in templates/cinder/config/10-cinder_wsgi.conf:

  WSGIDaemonProcess cinder-api display-name=cinder_wsgi group=cinder processes=4 threads=1 user=cinder

This patch does a temporary fix by adding the Cinder user to the tty group before starting the httpd service.

Ideally the cinder base image would be modified to include the cinder user from the start, so a review has been submitted to do precisely that: https://review.opendev.org/c/openstack/kolla/+/887370

### DIFF
--- a/templates/cinder/config/cinder-api-config.json
+++ b/templates/cinder/config/cinder-api-config.json
@@ -1,5 +1,5 @@
 {
-  "command": "/usr/sbin/httpd -DFOREGROUND",
+  "command": "usermod -a -G tty cinder; /usr/sbin/httpd -DFOREGROUND",
   "config_files": [
     {
       "source": "/var/lib/config-data/merged/cinder.conf.d",


steveb:
> Ideally the cinder base image would be modified to include the cinder user from the start, so a review has been submitted to do precisely that: https://review.opendev.org/c/openstack/kolla/+/887370

Adding this group in the image would need to be done here https://github.com/openstack-k8s-operators/tcib/blob/main/container-images/kolla/base/uid_gid_manage.sh#L39

steveb:
I tried this with custom ironic containers which had the ironic user added to the tty. The same issue occurs for me locally when I deploy with these images, so I don't think this approach will work

https://github.com/openstack-k8s-operators/tcib/compare/main...steveb:tcib:ttyironic

fmount:
I think this doesn't work properly [1] or at least it's not interpreted as it should.
Should we rework this part using the side container approach as done w/ nova? That should be a better long term solution I guess.

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_cinder-operator/223/pull-ci-openstack-k8s-operators-cinder-operator-main-cinder-operator-build-deploy-tempest/1674754162631380992/artifacts/cinder-operator-build-deploy-tempest/openstack-k8s-operators-gather/artifacts/pods/cinder-api-b884b58c6-6sgm6-containers.log

fmount:
@Akrog @abays @ASBishop I'm in the process of tagging both `Cinder` and `Glance` for dev preview as we agreed last week.
I realized that for `4.12+` we might want to land this fix, otherwise we keep seeing `cinder-api` crushing due to the issue described in the patch (permission denied on /dev/stdout).
However, I'm not sure this patch really addresses it, while the sidecar container solves pretty much the issue.
I have a working patch [1], where basically is the operator responsible to enforce the logging pattern to the container as part of its config (/dev/stdout vs logging file streamed in a sidecar). 
As a mid-term approach (to unblock tagging the repos) I'm wondering if it's something acceptable (we might require to revisit the logic in the controllers or we can easily follow up after tagging the repo).
I wanted to mention it here  to make sure you are able to check [1] and unblock the repos tagging during the next week.
FYI `Glance` is also blocked by `Cinder` tagging (we recently imported the `cinder-operator/api` due to the introduction of cinder as a glance backend).

[1] https://github.com/openstack-k8s-operators/cinder-operator/pull/236
[2] https://github.com/openstack-k8s-operators/glance-operator/pull/276

Akrog:
Abandon in favor of sidecar containers (PR#238)

PR #225, opened 2023-07-05
Backend Sample: Update multipath configuration

Akrog:
This patch updates the multipath sample configuration to set the "recheck_wwid" value to "yes".

This aligns with what we are recommending in the adoption documentation for the Cinder service.


ASBishop (approved):
/lgtm

PR #228, opened 2023-07-12
 Fix defaulting of Replicas spec

kajinamit:
Currently the default values are ignored because these specs are set to 0 by webhook. Update the type from `int32` to `*int32` to avoid that behavior and ensure the expected default values are used.

This also ensures Replicas spec takes values >= 0.


abays (commented):
Tested this with `install yamls` `make cinder_kuttl` and it passed

+1 from me

api/v1beta1/cinderapi_types.go
```diff
@@ -28,8 +28,9 @@ type CinderAPITemplate struct {
 
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=0
 	// Replicas - Cinder API Replicas
-	Replicas int32 `json:"replicas"`
+	Replicas *int32 `json:"replicas"`
```
ASBishop, line 33: As an old school C programmer in the distant past, I'm struggling to understand how this works. I gather the defaulting webhook will zero these fields, yielding the equivalent of a NULL pointer. Then the kubebuilder annotation will result in the replicas field being updated to point (because it's a pointer) to the default value, meaning the pointer is updated. Somehow kubebuilder knows how to set default values by reference AND by value (two different use cases).

Is my understanding anything close to correct?

pkg/cinderapi/deployment.go
```diff
@@ -91,7 +91,7 @@ func Deployment(
 			Selector: &metav1.LabelSelector{
 				MatchLabels: labels,
 			},
-			Replicas: &instance.Spec.Replicas,
+			Replicas: instance.Spec.Replicas,
```
ASBishop, line 94: This part confuses me (even the original code). According to the [StatefulSetSpec](https://pkg.go.dev/k8s.io/kubernetes/pkg/apis/apps#StatefulSetSpec), the Replicas field is an int32 (not a pointer), so I don't know how we can mix pointers and pointer references.

api/v1beta1/cinderapi_types.go
abays: > I gather the defaulting webhook will zero these fields

I'm not sure if the defaulting webhook sets the pointers to the zero-value of `nil` or if it just ignores them completely and thus they would remain `nil` as they would presumably already be.  But anyhow, then when they subsequently reach the kubebuilder logic, that logic indeed knows how to set the default value whether it's by reference or by value.  So your understanding is correct.

pkg/cinderapi/deployment.go
abays: For whatever reason, that documentation is not accurate.  If you look at the code in an interactive IDE, the imported libraries for `StatefulSet`, `Deployment`, etc, have an `*int32` as the data type for that field.

ASBishop: Ah, thanks for explaining this as a doc error. You've saved me from an eternity of staring into the abyss.

ASBishop (approved):
/lgtm

PR #229, opened 2023-07-14
Docs: Scheduler performance test results

Akrog:
We now recommend using a single scheduler as it avoids some inconsistency problems in the scheduling calculations and it's good enough for most of the deployments.

Since this is a deployment change, as we used to deploy 3 replicas, it's a good idea to document the change, briefly explain the reasons, and mention the expected performance of the service with a single replica.

This patch adds the top level explanation with the results of the tests and also a detailed explanation, in a different file to avoid cluttering, of the tests that were performed to obtain the results.


docs/scheduler.md
```diff
@@ -0,0 +1,43 @@
...
+operations, and sending the RPC call to the volume's queue in RabbitMQ, so our
+DB and RabbitMQ speeds are a big factor here.  If we exclude the RabbitMQ
+message handling and the DB access this goes down to less than 13ms per request.
+
+At the time of this writing for the create snapshot operation, which doesn't
+have DB requests, we only have the timing of the code without external services
+calls, which is around 1ms.
+
+Given this information it is our recommendation to run a single scheduler
+(`replicas: 1`) as it seems to be performant enough for most deployments. For
```
ASBishop, line 33: This recommendation should be reflected in the kubebuilder min/max replicas. If 1 replica is sufficient for "most" deployments, then should we set the max replicas allowed to 2? 3?

docs/scheduler-perf/scheduler-perf.md
```diff
@@ -0,0 +1,856 @@
...
+          [Install]
+          WantedBy=local-fs-pre.target
+```
+
+After the node was rebooted the usual steps were followed to deploy things:
+```
+cd install_yamls/devsetup
+eval $(crc oc-env)
+# Link kubectl to oc since we don't have kubectl installed
+ln -s oc `dirname \`which oc\``/kubectl
```
ASBishop, line 272: I suppose there's no harm including these lines, but they are extraneous. Nothing described in this document uses the kubectl command.

docs/scheduler-perf/scheduler-perf.md
```diff
@@ -0,0 +1,856 @@
...
+      passwordSelectors:
+        database: CinderDatabasePassword
+        service: CinderPassword
+      preserveJobs: false
+      rabbitMqClusterName: rabbitmq
+      secret: osp-secret
+      serviceUser: cinder
+```
+
+The cinder volume service was setup to not start automatically to allow us to
```
ASBishop, line 371: It took me a moment to figure out what you meant, and I now see that the statement is true because of what's on L349. I think this could be made more explicit (but I don't have a specific recommendation on how to do that).

docs/scheduler-perf/scheduler-perf.md
```diff
@@ -0,0 +1,856 @@
...
+modify the code and reduce the impact on the node of the volume creation part:
+```bash
+$ oc exec -it cinder-volume-lvm-iscsi-0 -- /bin/bash
```
ASBishop, line 374: Why not use "oc rsh"?

fwiw, I pip installed openstackclient on my system so that I have native access to the command.

docs/scheduler-perf/scheduler-perf.md
```diff
@@ -0,0 +1,856 @@
...
+$ /usr/local/bin/kolla_set_configs
+$ /usr/local/bin/kolla_start
+```
+
+We leave that terminal alone to let the cinder volume service run.
+
+From the host we now modify the quotas to remove the limit on volumes and
+gigabytes for the tests:
+```bash
+$ oc exec -it openstackclient -- openstack quota set --gigabytes=-1 --volumes=-1 admin
```
ASBishop, line 398: oc rsh?

docs/scheduler-perf/scheduler-perf.md
```diff
@@ -0,0 +1,856 @@
...
+  port:
+    targetPort: management
+  to:
+    kind: Service
+    name: rabbitmq
+    weight: 100
+  wildcardPolicy: None
+```
+
+We get the cinder credentials (<user>:<password>) for RabbmitMQ by looking at
```
ASBishop, line 427: -1: This didn't render correctly, because markdown doesn't preserve the angle brackets and so they end up looking like (bogus) HTML tags. Swapping them with HTML character codes [1] worked for me.

[1] https://www.coderepublics.com/Reference/html-character-entities.php

docs/scheduler-perf/scheduler-perf.md
```diff
@@ -0,0 +1,856 @@
...
+```bash
+$ sudo dnf -y install httpd-tools
+```
+
+And we finally create the JSON for the request, get the credentials for the
+tests, and run the volume creation 4000 request with a concurrency of 20 using
+the `ab` command:
+```bash
+$ echo '{"volume": {"size": 1, "consistencygroup_id": null, "snapshot_id": null, "name": "40", "description": null, "volume_type": null, "availability_zone": null, "metadata": {}, "imageRef": null, "source_volid": null, "backup_id": null}}' > create-vol.json
+$ token_id=`oc exec -t openstackclient -- openstack token issue -f value -c id`
```
ASBishop, line 451: oc rsh? though maybe that generates extra garbage when it explains which container within the pod it chooses.

ASBishop:
This is very cool! The results are favorable (one c-sch looks to be sufficient!), and the data collection and analysis is very impressive.

Unrelated to this PR, it suggests we should do some global cleanup of our docs and top level README. Perhaps we should relocate the important stuff into the docs directory, and have the top level README point you there. I'm ambivalent whether we should publish the docs on openstack-k8s-operators.github.io, but it's something to consider.

ASBishop (changes requested):

docs/scheduler.md
Akrog: I still feel the same as in PR#205, that we should not limit the max replicas unless we know it's something that won't work (like for the volume service).

docs/scheduler-perf/scheduler-perf.md
Akrog: Makefile targets use it and won't work if it's not there.
Although this code I added a while back should take care of it... https://github.com/openstack-k8s-operators/cinder-operator/blob/53e4d6a340a78012f2d2a09d1b37d2bbbd17f48a/Makefile#L78

So maybe in my tests I forgot to do the `eval $(crc oc-env)` part before calling `make`...

Akrog: Silly me, I added that to the cinder operator, not the install-yamls repository, so that's why it is necessary.

ASBishop (approved):
/lgtm

PR #233, opened 2023-07-21
Use central renovate config

fmount:
As done for other operators, this change switches renovate to a central config.


abays (approved):
/lgtm

PR #236, opened 2023-07-22
sidecar logging for cinder-api

fmount:
cinder-api is failing in ocp 4.12+ due to the recent permission denied error on `/dev/stdout`.
For this reason it might be a good idea  introduce a sidecar container to stream logs.
This is consistent with the k8s official pattern [1].

[1] https://kubernetes.io/docs/concepts/cluster-administration/logging

### DIFF
--- a/templates/cinder/config/00-config.conf
+++ b/templates/cinder/config/00-config.conf
@@ -23,10 +23,6 @@ osapi_volume_workers = 4
 control_exchange = openstack
 api_paste_config = /etc/cinder/api-paste.ini
 
-# For containers we want to log to stdout instead of the default stderr, as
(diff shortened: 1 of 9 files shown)


templates/cinder/config/00-config.conf
```diff
@@ -23,10 +23,6 @@ osapi_volume_workers = 4
 control_exchange = openstack
 api_paste_config = /etc/cinder/api-paste.ini
 
-# For containers we want to log to stdout instead of the default stderr, as
```
fmount, line 26: We don't need this anymore as a "global" option because we can enforce it per-service

fmount:
I suspect I'm still not on the right path with this, closing for now.

PR #238, opened 2023-07-22
sidecar logging for cinder-api

fmount:
Due to the recent permission denied error on `/dev/stdout`, we're introducing a `sidecar container` to stream logs
in cinder-api. 
This is also consistent with the k8s official pattern [1].

[1] https://kubernetes.io/docs/concepts/cluster-administration/logging


pkg/cinderapi/deployment.go
```diff
@@ -100,6 +100,24 @@ func Deployment(
...
+							Args:  []string{"-c", "tail -n+1 -F " + LogPath},
+							Image: instance.Spec.ContainerImage,
+							SecurityContext: &corev1.SecurityContext{
+								RunAsUser: &runAsUser,
+							},
+							Env:            env.MergeEnvs([]corev1.EnvVar{}, envVars),
+							VolumeMounts:   []corev1.VolumeMount{GetLogVolumeMount()},
+							Resources:      instance.Spec.Resources,
+							ReadinessProbe: readinessProbe,
+							LivenessProbe:  livenessProbe,
```
Akrog, line 119: In my opinion this container should not have a readiness or liveness probe. Having the same as the API would double the number of requests to the API and may unnecessarily restart the logging pod.

controllers/cinderapi_controller.go
```diff
@@ -779,6 +779,13 @@ func (r *CinderAPIReconciler) generateServiceConfigs(
 	}
 	customData[cinder.CustomServiceConfigSecretsFileName] = customSecrets
 
+	// Inject Default logging: /dev/stdout doesn't work for cinder-api and a
+	// sidecar container is used to stream the logs to the defined LogPath.
+	// The following line makes the cinder-operator responsible to append to the
+	// ServiceConfig the logging line by default.
+	customData[cinder.CustomServiceConfigFileName] = fmt.Sprintf("%s%s%s\n",
+		string(cinderSecret.Data[cinder.CustomServiceConfigFileName]), cinder.LogSnippet, cinderapi.LogPath)
```
Akrog, line 787: I don't remember, but does our secret always end in `\n` or should we add it to the `Sprintf`
```
   fmt.Sprintf("%s\n%s%s\n",
```

Akrog (changes requested):
One question and one minor change. Looks good otherwise.

pkg/cinderapi/deployment.go
fmount: Done

controllers/cinderapi_controller.go
fmount: Right, I don't think it's present when we build the `CustomServiceConfig` from the Spec:

```
customData[cinder.CustomServiceConfigFileName] = instance.Spec.CustomServiceConfig
```
no one ensures that a \n is always there.
I added a '\n' at the beginning of Sprintf, so we can always make sure to have it (and it doesn't hurt if it's present)

fmount: @Akrog ok I think this part is Done, and kuttl tests are also updated.
Let me know if this is good for you and we'll land this before tagging the Cinder repo.

Akrog: Not really loving that we are writing config data inside the user snippet file (`02-config.conf`), as I think that should be an exact match of what the user wrote, but we can change that afterwards, as it's not a user facing change.

Akrog (approved):
/lgtm

PR #244, opened 2023-08-02
 Require databaseHostname and transportURLSecret

kajinamit:
These spec fields are required to deploy cinder services.

Although we currently do not intend to support direct usage of sub CRs, requiring these fields allows us to catch any bug in cinder controller.


Akrog (approved):
/lgtm

PR #245, opened 2023-08-03
Customize 00-default.conf for better logging

fmount:
This patch represents a follow-up of #238 where a sidecar container for `cinder-api` logging purposes has been introduced.
This PR represents a cleaner way to handle the override for `cinder-api`, and it relies on the existing `00-config.conf` template to inject the right `log_file` path.
In general, the top-level controller generates a default `ConfigFile`, which, in most cases (right now for all the services except `cinder-api`), is inherited and set in the related k8s `Secret` of a given service. However, the `cinder-api` controller implements the logic of properly handle overrides via `templateParameters`.
This logic is at least shared between the top-level and the `cinder-api` controllers, and, in the future, it can be extended to support multiple overrides on additional templates through the same mechanism, hence a `common.go` has been created and can be used to make more "deduplication" as long as common code is detected.


controllers/common.go
```diff
@@ -0,0 +1,74 @@
...
+	"fmt"
+	"github.com/openstack-k8s-operators/cinder-operator/pkg/cinder"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/env"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func GenerateConfigsGeneric(
```
fmount, line 30: @ASBishop I'm pretty sure that going deeper in the `cinder_controller*` logic we'll find that we can "deduplicate" more that just this function. If it's ok with you I would leave some code cleanup to a  follow up PR, but here the idea is to make sure that we call a common function that allows the service to customize the existing template (e.g., for logging purposes) instead of just inheriting it (which is the default behavior).

controllers/cinderapi_controller.go
```diff
@@ -779,25 +777,48 @@ func (r *CinderAPIReconciler) generateServiceConfigs(
 	}
 	customData[cinder.CustomServiceConfigSecretsFileName] = customSecrets
 
-	// Inject Default logging: /dev/stdout doesn't work for cinder-api and a
-	// sidecar container is used to stream the logs to the defined LogPath.
-	// The following line makes the cinder-operator responsible to append to the
-	// ServiceConfig the logging line by default.
-	customData[cinder.CustomServiceConfigFileName] = fmt.Sprintf("%s\n%s%s\n",
```
fmount, line 786: @Akrog instead of injecting an arbitrary snippet for Logging in `CustomServiceConfigFileName`, we're making the service able to use `templateParameters` and customize `00-config.conf` independently instead of inheriting it like the other services.
This is the basically the price to pay to make the service able to "diverge" a bit in its config according to the needs.
However, almost everything is still common among the services (which is the reason the `EnsureSecret` function has been called from `common.go`), but here we have the ability to specify a different `LogPath` if required without "hacking" any existing config file.

pkg/cinder/const.go
```diff
@@ -65,8 +65,8 @@ const (
 	// Cinder is the global ServiceType that refers to all the components deployed
 	// by the cinder operator
 	Cinder storage.PropagationType = "Cinder"
-	//LogSnippet -
-	LogSnippet = "[DEFAULT]\nlog_file="
```
fmount, line 69: This seemed more a hack that helped for the cinder repo tagging for dev preview.
I removed the `LogSnippet` with this patch as services should be able to have this information in the `00-config.conf` template itself and at least they should be able to provide the path value.

Akrog (commented):
To be honest, I am not loving this whole templating approach here.

It looks like we are growing our code base when we could just be using a single additional configuration file with the parameters that the specific service wants to override, for example a config file called `00-config-log.conf` or `04-forced-conf.conf`.

This code could come from a file just like the `00-config.conf` that we have.

ASBishop:
I confess that I've put off reviewing this PR in order to focus on something else. I haven't studied the changes in depth, but it does feel like quite a bit of code that might represent a step sideways. It does help address how to fix cinder-api's log path, but do we expect it to pay dividends for similar situations that might arise in the future?

Let me offer an alternative approach. Basically, it's to use a small hook I developed for the cinder-volume service that lets you "extend" the service's configuration. The idea is we'd leave 00-config.conf alone, and add something to cinder-api's 02-config.conf (aka, the CustomServiceConfigFileName). That way, c-api's service specific config would override the global setting.

Take a look at cinder-volume's code at [1] and [2].

[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/controllers/cindervolume_controller.go#L555
[2] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/controllers/cindervolume_controller.go#L624

cinder-api would implement its own extendCustomServiceConfig() function that would only need to do something like this:

```
func extendCustomServiceConfig(customServiceConfig string) string {
	// WSGI applications like cinder-api need to log to stdout.
	return "[DEFAULT]\nlog_file=/dev/stdout\n" + customServiceConfig
}
```

ASBishop:
Gorka and I chatted, and agree to not use the extendCustomServiceConfig() method I described above, but instead see about revising the list of xx-config.conf files to support a new one for service specific defaults, which would be loaded immediately after the 00-config file (global defaults) and the "global user defined defaults" (currently the 01 file).

fmount:
> Gorka and I chatted, and agree to not use the extendCustomServiceConfig() method I described above, but instead see about revising the list of xx-config.conf files to support a new one for service specific defaults, which would be loaded immediately after the 00-config file (global defaults) and the "global user defined defaults" (currently the 01 file).

Ack and thanks a lot for the feedback.
I see you already have a path forward for this kind of change and I'm ok about
closing this PR in favor of something else that would fit better the purpose of
enforcing specific defaults.
I understand you think that the codebase here is growing, however, let me explain
a few reasons that led me to this kind of patch:

1. I tried to add a specific `XY-default-log.conf` containing only the `LogSnippet`
and the `LogPath` (I think it's better to have a `LogSnippet` in a sort of template
file instead of being a string that is assigned or concatenated in `CustomServiceConfig`).
However, this approach didn't work for me: overriding the `[DEFAULT]` section somehow
made the `cinder-api` `POD` lose the information of `00-config.conf`, including the 
database connection, and resulted in a `CrushLoopBack`. Feel free to revisit this
strategy, maybe I made (logical) mistakes and I reached wrong assumptions.

2. It's true and efficient that `00-config.conf` is inherited for all the services and
specific sections can be overridden via `CustomServiceConfig`, but the cost of
centralizing the template processing in a common function that can be fed with
different `templateParameters` variables seems reasonable for earning flexibility
in case a parameter should diverge for a service (or a subset of services). Of
course, if no input is passed, the old behavior is preserved, and services continue
to inherit the default `00-config.conf` file.
An additional reason, on this front, is that I thought it was better to reuse the
existing template instead of introducing yet another file only for this purpose: your
proposal about identifying a list of defaults that should be loaded right after
`00-config` is fine I guess, but I'm wondering if this brings us back to the previous
point (just with more parameters instead of logging only).

3. Having a `common.go` can fit the purpose of "deduplicating" code: I didn't go in
depth into this topic but I suspect we're going to find more logic (e.g., in the 
networking side [1][2][3][4]) that can be in common. Generating the config that is
mostly the same (except for the overrides) seems to be in line with this topic, and
helped this log patch because by just providing a few lines in the existing template
and feeding it w/ diff input made the trick (see `templatesParameter["LogPath"]`).

4. Finally, I find cleaner (from a coding/maintenance point of view), establishing a
pattern in the way we manage golang templates instead of trying to fix it by adding
more logic to the operator: to be honest I like the idea of producing a template that
reflects a given input, and while the operator should be responsible to generate that
input according to a given logic, we can defer to golang templating the generation of
the file itself [5]. This also allows each service to choose whether it makes sense to
inherit a `00-config` (or any other config file generated by the umbrella controller), or
override it. The same pattern can be applied to any other file generated by the main
controller.

Being that said, I'm glad you both shared your feedback and your thoughts, and
I'm going to close it in favor of the strategy you think it's best for the operator.

[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/controllers/cinderscheduler_controller.go#L352
[2] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/controllers/cinderbackup_controller.go#L353
[3] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/controllers/cinderapi_controller.go#L556
[4] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/controllers/cindervolume_controller.go#L354
[5] https://pkg.go.dev/text/template

PR #250, opened 2023-08-15
Support service specific config snippets

ASBishop:
Add support for a new service specific config snippet, which is loaded after the global snippet and before the "custom" global snippet supplied by the user. This adds the ability for the operator to override certain global settings on a per-service basis.

The immediate use case is to provide a cleaner way to override the logging configuration for the cinder-api service.

The [backend_defaults] section in the global snippet has been moved into a new cinder-volume service specific snippet.

### DIFF
--- a/pkg/cinder/const.go
+++ b/pkg/cinder/const.go
@@ -33,12 +33,14 @@ const (
 
 	// DefaultsConfigFileName -
 	DefaultsConfigFileName = "00-config.conf"
+	// ServiceConfigFileName -
+	ServiceConfigFileName = "01-config.conf"
 	// CustomConfigFileName -
-	CustomConfigFileName = "01-config.conf"
+	CustomConfigFileName = "02-config.conf"
 	// CustomServiceConfigFileName -
-	CustomServiceConfigFileName = "02-config.conf"
+	CustomServiceConfigFileName = "03-config.conf"
 	// CustomServiceConfigSecretsFileName -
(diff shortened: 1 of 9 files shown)


templates/cindervolume/config/01-config.conf
Akrog, line 1: Should we add one for the backup service as well?
The backup service is the one that uses the same parameter in the defaults section: https://github.com/openstack-k8s-operators/cinder-operator/blob/main/templates/cinder/config/00-config.conf#L36

fmount:
+1 on having `01-config.conf` independently configured by each service

pkg/cinder/const.go
```diff
@@ -33,12 +33,14 @@ const (
 
 	// DefaultsConfigFileName -
 	DefaultsConfigFileName = "00-config.conf"
+	// ServiceConfigFileName -
+	ServiceConfigFileName = "01-config.conf"
```
Akrog, line 37: It may be a good idea to rename these files at some point, not in this PR, so that the names are self explanatory:

```
00-global-defaults.conf
01-service-defaults.conf
02-global-custom.conf
03-service-custom.conf
04-service-custom-secrets.conf
```

ASBishop: I agree the files should be renamed. There was a reason why I originally chose generic names (they don't even mention "cinder"), but time has passed and I think the descriptive names make more sense. I can update this PR next week (I'm on PTO now), or submit another PR later.

Akrog (approved):

PR #251, opened 2023-08-15
Stop using KOLLA_CONFIG_FILE environment variable

ASBishop:
Use a volume mount instead of overriding KOLLA_CONFIG_FILE for each service.

### DIFF
--- a/pkg/cinder/const.go
+++ b/pkg/cinder/const.go
@@ -45,9 +45,6 @@ const (
 	// CinderInternalPort -
 	CinderInternalPort int32 = 8776
 
-	// KollaConfigDbSync -
-	KollaConfigDbSync = "/var/lib/config-data/merged/db-sync-config.json"
-
 	// CinderExtraVolTypeUndefined can be used to label an extraMount which
 	// is not associated with a specific backend
 	CinderExtraVolTypeUndefined storage.ExtraVolType = "Undefined"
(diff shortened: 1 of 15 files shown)


kajinamit:
Thanks for working on this !

According to the error in kuttl tests I think you also need to update the assertion file.

```
resource Deployment:cinder-kuttl-tests/cinder-api: .spec.template.spec.containers.volumeMounts: slice length mismatch: 6 != 7
```

ASBishop:
This is odd, as I don't see this affecting similar patches in other operators. I can't reproduce the problem locally, and if I try adding the additional mount it then fails when I run it locally.

I try adding and see what happens.

kajinamit:
> This is odd, as I don't see this affecting similar patches in other operators. I can't reproduce the problem locally, and if I try adding the additional mount it then fails when I run it locally.


Unfortunately we currently have quite inconsistent test coverage/assertion levels in kuttl tests of each operator. In some operators the kuttl tests checkes quite detailed specs of individual deployment while other operators checks only existence and a few core keys such as owner reference or finalizers. That's why we haven't seen the same errors in other operators we've updated so far.

# I'm unsure why the error does not occur in your local env, though

kajinamit:
The change looks good to me. I'm not adding my lgtm label in case you want someone from the storage team to review this change, but lmk if you want me to merge this PR.

PR #255, opened 2023-08-29
Including cinder-api/backup/scheduler/volume in the deploy step of kuttl run

frenzyfriday:


abays (approved):
Tested in my environment and KUTTL passed

/lgtm

PR #257, opened 2023-09-01
Update cinder.conf defaults

Akrog:
This patch updates the global defaults in `00-config.conf` to:

- Check service token roles in the KeyStone middleware.
- Allow cloning of Glance image volumes when Glance is using Cinder as a backend.


ASBishop (approved):
/lgtm

PR #258, opened 2023-09-02
switch to structured logging

pinikomarov:
This automatically adds additional fields like reconcile_id etc.. from the controller context.

before :

`2023-05-18T01:53:14+03:00 INFO  controllers.KeystoneAPI Reconciled Service init successfully`

after:

`2023-10-04T16:06:31+03:00       INFO    Controllers.CinderAPI   Reconciled Service 'cinder-api' successfully    {"controller": "cinderapi", "controllerGroup": "cinder.openstack.org", "controllerKind": "CinderAPI", "CinderAPI": {"name":"cinder-api","namespace":"openstack"}, "namespace": "openstack", "name": "cinder-api", "reconcileID": "09fed477-8ad5-4186-8b20-790f409de78a"}
`

*by using per reconcile function respective logger objects, the intention is to lay the ground for parallel reconciliation in future and avoid race conditions as ctx objects are reconcile run specific.

### DIFF
--- a/controllers/cinder_controller.go
+++ b/controllers/cinder_controller.go
@@ -70,11 +70,6 @@ func (r *CinderReconciler) GetKClient() kubernetes.Interface {
 	return r.Kclient
 }
 
-// GetLogger -
-func (r *CinderReconciler) GetLogger() logr.Logger {
-	return r.Log
-}
-
 // GetScheme -
 func (r *CinderReconciler) GetScheme() *runtime.Scheme {
 	return r.Scheme
@@ -84,8 +79,12 @@ func (r *CinderReconciler) GetScheme() *runtime.Scheme {
 type CinderReconciler struct {
 	client.Client
 	Kclient kubernetes.Interface
-	Log     logr.Logger
 	Scheme  *runtime.Scheme
+}
+
+// GetLogger returns a logger object with a logging prefix of "controller.name" and additional controller context fields
+func (r *CinderReconciler) GetLogger(ctx context.Context) logr.Logger {
+	return log.FromContext(ctx).WithName("Controllers").WithName("Cinder")
 }
 
 // +kubebuilder:rbac:groups=cinder.openstack.org,resources=cinders,verbs=get;list;watch;create;update;patch;delete
@@ -122,7 +121,7 @@ type CinderReconciler struct {
 
 // Reconcile -
 func (r *CinderReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, _err error) {
-	_ = log.FromContext(ctx)
+	Log := r.GetLogger(ctx)
 
 	// Fetch the Cinder instance
 	instance := &cinderv1beta1.Cinder{}
@@ -143,7 +142,7 @@ func (r *CinderReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 		r.Client,
 		r.Kclient,
 		r.Scheme,
-		r.Log,
+		Log,
 	)
 	if err != nil {
 		return ctrl.Result{}, err
@@ -218,7 +217,7 @@ func (r *CinderReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 }
 
 // SetupWithManager sets up the controller with the Manager.
-func (r *CinderReconciler) SetupWithManager(mgr ctrl.Manager) error {
+func (r *CinderReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
 	// transportURLSecretFn - Watch for changes made to the secret associated with the RabbitMQ
 	// TransportURL created and used by Cinder CRs.  Watch functions return a list of namespace-scoped
 	// CRs that then get fed  to the reconciler.  Hence, in this case, we need to know the name of the
@@ -233,13 +232,15 @@ func (r *CinderReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	transportURLSecretFn := func(o client.Object) []reconcile.Request {
 		result := []reconcile.Request{}
 
+		Log := r.GetLogger(ctx)
+
 		// get all Cinder CRs
 		cinders := &cinderv1beta1.CinderList{}
 		listOpts := []client.ListOption{
 			client.InNamespace(o.GetNamespace()),
 		}
-		if err := r.Client.List(context.Background(), cinders, listOpts...); err != nil {
-			r.Log.Error(err, "Unable to retrieve Cinder CRs %v")
+		if err := r.Client.List(ctx, cinders, listOpts...); err != nil {
+			Log.Error(err, "Unable to retrieve Cinder CRs %v")
 			return nil
 		}
 
@@ -252,7 +253,7 @@ func (r *CinderReconciler) SetupWithManager(mgr ctrl.Manager) error {
 							Namespace: o.GetNamespace(),
 							Name:      cr.Name,
 						}
-						r.Log.Info(fmt.Sprintf("TransportURL Secret %s belongs to TransportURL belonging to Cinder CR %s", o.GetName(), cr.Name))
+						Log.Info(fmt.Sprintf("TransportURL Secret %s belongs to TransportURL belonging to Cinder CR %s", o.GetName(), cr.Name))
 						result = append(result, reconcile.Request{NamespacedName: name})
 					}
 				}
@@ -265,6 +266,8 @@ func (r *CinderReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}
 
 	memcachedFn := func(o client.Object) []reconcile.Request {
+		Log := r.GetLogger(ctx)
+
 		result := []reconcile.Request{}
 
 		// get all Cinder CRs
@@ -273,7 +276,7 @@ func (r *CinderReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			client.InNamespace(o.GetNamespace()),
 		}
 		if err := r.Client.List(context.Background(), cinders, listOpts...); err != nil {
-			r.Log.Error(err, "Unable to retrieve Cinder CRs %w")
+			Log.Error(err, "Unable to retrieve Cinder CRs %w")
 			return nil
 		}
 
@@ -283,7 +286,7 @@ func (r *CinderReconciler) SetupWithManager(mgr ctrl.Manager) error {
 					Namespace: o.GetNamespace(),
 					Name:      cr.Name,
 				}
-				r.Log.Info(fmt.Sprintf("Memcached %s is used by Cinder CR %s", o.GetName(), cr.Name))
+				Log.Info(fmt.Sprintf("Memcached %s is used by Cinder CR %s", o.GetName(), cr.Name))
 				result = append(result, reconcile.Request{NamespacedName: name})
 			}
 		}
@@ -316,7 +319,9 @@ func (r *CinderReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 func (r *CinderReconciler) reconcileDelete(ctx context.Context, instance *cinderv1beta1.Cinder, helper *helper.Helper) (ctrl.Result, error) {
-	r.Log.Info(fmt.Sprintf("Reconciling Service '%s' delete", instance.Name))
+	Log := r.GetLogger(ctx)
+
+	Log.Info(fmt.Sprintf("Reconciling Service '%s' delete", instance.Name))
 
 	// remove db finalizer first
 	db, err := mariadbv1.GetDatabaseByName(ctx, helper, instance.Name)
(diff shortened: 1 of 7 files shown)


controllers/cinderscheduler_controller.go
```diff
@@ -87,7 +79,7 @@ type CinderSchedulerReconciler struct {
 
 // Reconcile -
 func (r *CinderSchedulerReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, _err error) {
-	_ = log.FromContext(ctx)
+	log := GetLog(ctx, "CinderBackup")
```
Akrog, line 82: -1: All calls in this file should be using `CinderScheduler` or `CinderSch` whatever we prefer.

Akrog (changes requested):

controllers/cinderscheduler_controller.go
pinikomarov: right , correcting

controllers/cinder_controller.go
```diff
@@ -223,6 +222,9 @@ func (r *CinderReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	// matches the name of an existing Cinder CR.  In that case changes to that secret would trigger
 	// reconciliation for a Cinder CR that does not need it.
 	//
+
+	log := GetLog(ctx, "Cinder")
```
ASBishop, line 226: Is there any way to persist this, similar to how the previous code used r.log? My concern is that every function that wants to log something will need to repeat L226 in order to obtain its own 'log' object.

controllers/cinder_controller.go
```diff
@@ -278,7 +280,8 @@ func (r *CinderReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 func (r *CinderReconciler) reconcileDelete(ctx context.Context, instance *cinderv1beta1.Cinder, helper *helper.Helper) (ctrl.Result, error) {
-	r.Log.Info(fmt.Sprintf("Reconciling Service '%s' delete", instance.Name))
+	log := GetLog(ctx, "Cinder")
```
ASBishop, line 283: This is an example of what I referred to in my previous comment.

controllers/cinder_controller.go
ASBishop: I'm experimenting with an alternative approach, and might submit my own PR just as a PoC.

pinikomarov: Hi Alan, we're trying the drop the use of the global logger in the struct,
_We discussed it here :
It's to do with parallel reconcile use case and making room for differentiating each reconcile_id if that happens.
https://github.com/openstack-k8s-operators/keystone-operator/pull/220#issuecomment-1483117420_

Also the Gibbi believes that the pattern of of using the r.log from the global struct 
to pass request specific info between different reconcile calls is not a good pattern.
_Gibbi commented on the r.log usage here : 
https://github.com/openstack-k8s-operators/nova-operator/pull/496#discussion_r1307394210_

@gibizer fyi ^^

about placing the getLog.. func in the funcs module - cinder.GetLoggerWithContex , I'm good with that either way

gibizer: I agree to remove the repetition with a wrapper (or with changing the logger in the ctx object once)

pinikomarov: As I see it we have to options here :
  
// Option 1:

// Per controller wrapper function that wraps the GetLog
func getLogWrapper() func(context.Context, string) logr.Logger{
	return GetLog(ctx, "HeatAPI")
}

//usage
func (...) Reconcile...
	log := getLogWrapper()


// OR

// Option 2 ,at the head of each controller_.go we have a non exported local func with a statically named controller:
func getLog(ctx context.Context) logr.Logger {
	return log.FromContext(ctx).WithName("Controllers").WithName("Heat")
}

//usage
func (...) Reconcile...
	log := getLog(ctx)


// Either way I think about implementation, the simplest use case to seems to me is option 2.

WDYT ?

gibizer: Option 2 looks good to me

ASBishop: I also prefer option 2.

pinikomarov: Another issue following that is that the getlog func signitures need to be unique across the controllers since they're all in the same "package controllers".

For that I think this can work : 
each controller will get a unique get log sign.: 
```go 
func cinderApiLog(ctx context.Context) logr.Logger ..
```
//usage : 
```go
func (..) Reconcile..ctrl.Result, _err error) {
	log := cinderApiLog(ctx)
```

How's that ?

gibizer: fine by me

controllers/cinder_controller.go
```diff
@@ -67,9 +67,9 @@ func (r *CinderReconciler) GetKClient() kubernetes.Interface {
 	return r.Kclient
 }
 
-// GetLogger -
-func (r *CinderReconciler) GetLogger() logr.Logger {
-	return r.Log
+// getLog returns a logger object with a logging prefix of "conrollers.name" and aditional controller context fields
+func cinderLog(ctx context.Context) logr.Logger {
```
Akrog, line 71: -1: Please make this a struct method like it was before, all you need to do is add the new method argument.

```
func (r *CinderReconciler) GetLogger(ctx context.Context) logr.Logger {
```

That way the name remains the same regardless of the controller we are in and we'll always do `log := r.GetLogger(ctx)`

Akrog (changes requested):
Thanks for all the hard work to improve our logging :-)

Akrog (commented):
I think the code is ready to be merged, could you squash all the commits into 1 please?
And rebase it, because it's in conflict.

controllers/cinder_controller.go
```diff
@@ -67,9 +67,9 @@ func (r *CinderReconciler) GetKClient() kubernetes.Interface {
 	return r.Kclient
 }
 
-// GetLogger -
-func (r *CinderReconciler) GetLogger() logr.Logger {
-	return r.Log
+// getLog returns a logger object with a logging prefix of "conrollers.name" and aditional controller context fields
+func (r *CinderReconciler) getLog(ctx context.Context) logr.Logger {
```
ASBishop, line 71: Every controller has a set of GetXXX() functions at the top, which have a unique signature based on their reconciler structure. I think we should retain that pattern, and not rename each controller's GetLogger() function to getLog().

In other words, can we keep the GetLogger() name?

controllers/cinder_controller.go
```diff
@@ -138,7 +137,7 @@ func (r *CinderReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 		r.Client,
 		r.Kclient,
 		r.Scheme,
-		r.Log,
+		log,
```
ASBishop, line 140: I really like the idea of adding the logger function to the helper !!!

But can we call it Log (capital 'L')? See how it will be referenced in a comment below.

controllers/cinder_controller.go
```diff
@@ -278,7 +280,8 @@ func (r *CinderReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 func (r *CinderReconciler) reconcileDelete(ctx context.Context, instance *cinderv1beta1.Cinder, helper *helper.Helper) (ctrl.Result, error) {
-	r.Log.Info(fmt.Sprintf("Reconciling Service '%s' delete", instance.Name))
+	log := r.getLog(ctx)
```
ASBishop, line 283: Why not use the logger that's available in the helper?

```
    helper.Log.Info(...)
```

controllers/cinder_controller.go
```diff
@@ -309,7 +312,8 @@ func (r *CinderReconciler) reconcileInit(
 	serviceLabels map[string]string,
 	serviceAnnotations map[string]string,
 ) (ctrl.Result, error) {
-	r.Log.Info(fmt.Sprintf("Reconciling Service '%s' init", instance.Name))
+	log := r.getLog(ctx)
+	log.Info(fmt.Sprintf("Reconciling Service '%s' init", instance.Name))
```
ASBishop, line 316: Same as before. Just replace "r.Log()" with "helper.Log()"

ASBishop (changes requested):
This PR is moving in the right direction, but I want us to consider another update.

I also agree with Gorka's note about rebasing and squashing the commits.

controllers/cinder_controller.go
gibizer: For me the Helper is a dumping ground of unrelated utility functions. See https://wiki.c2.com/?AbuseOfUtilityClasses So my ultimate goal is not to add things to it but actually dissolve it in the long run.

gibizer: I had a stab at having a separate logger package https://github.com/openstack-k8s-operators/lib-common/pull/125  but it turned out that we already get a proper logger via the ctx anyway so that package would be fairly empty.

ASBishop: So where does that leave us? It seems reasonable for every controller to have its own logger function, and the question is how to provide easy access to that function. I agree the Helper shouldn't be a dumping ground, but if we avoid using it then we'll need some other way to forward.

gibizer: the `ctx` is already passed around and the `ctx` already contains a a logger to use. So the `r.getLog(ctx)` is good for me.

gibizer: I.e for me the difference between ` helper.Log.Info(...)` an `r.getLog(ctx).Info()` is that ctx is the framework's way to pass a logger to us and we already have the context in most places, so pulling the logger out of context and adding it to the helper seems like an unnecessary complication to me.

controllers/cinder_controller.go
```diff
@@ -116,7 +115,7 @@ type CinderReconciler struct {
 
 // Reconcile -
 func (r *CinderReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, _err error) {
-	_ = log.FromContext(ctx)
+	Log := r.GetLogger(ctx)
```
ASBishop, line 118: I still wish it was possible to pass the logger function down the call stack in order to minimize the number of places that have to call this function. But maybe I just need to get used to a new (to me) convention.

controllers/cinder_controller.go
```diff
@@ -137,7 +136,7 @@ func (r *CinderReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 		r.Client,
 		r.Kclient,
 		r.Scheme,
-		r.Log,
+		Log,
```
ASBishop, line 139: @gibizer I recall you saying you'd like to reduce (eliminate?) our use of helpers. With this PR, the Log element in the helper is no longer used by the controller, but we have to supply one in order create the helper. It looks like we have an opportunity to remove the Log from the helper (since nobody will use it), but I don't know how to coordinate doing that across all the repos that use the current helper. Any thoughts on that?

ASBishop (commented):
+1 from me, though I'd like to see feedback from @Akrog .

Also see my comment on how we might take the opportunity to remove the Log element from helpers.

controllers/cinder_controller.go
```diff
@@ -66,9 +66,9 @@ func (r *CinderReconciler) GetKClient() kubernetes.Interface {
 	return r.Kclient
 }
 
-// GetLogger -
-func (r *CinderReconciler) GetLogger() logr.Logger {
-	return r.Log
+// GetLogger returns a logger object with a logging prefix of "conrollers.name" and aditional controller context fields
```
olliewalsh, line 69: nit: typos `conrollers.name` and `aditional`, copied multiple times

olliewalsh (changes requested):

controllers/cinder_controller.go
Akrog: I agree, we could have had a PR to remove the parameter in lib-common, and then bump its pinning here in this same PR so we could remove the passing of the parameter in one go.

With the current approach we need to:

- Merge this PR
- Remove the code from lib-common that requires the logger argument
- Update the pinning in all the operators at the same time we remove the passing of the argument.

Our current approach seems more painful, but for the shake of getting things done and keep moving we can merge it.

Akrog (approved):
/lgtm

pinikomarov:
recheck

controllers/cinder_controller.go
```diff
@@ -231,13 +230,15 @@ func (r *CinderReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		result := []reconcile.Request{}
 
+		Log := r.GetLogger(ctx)
+
 		// get all Cinder CRs
 		cinders := &cinderv1beta1.CinderList{}
 		listOpts := []client.ListOption{
 			client.InNamespace(o.GetNamespace()),
 		}
 		if err := r.Client.List(context.Background(), cinders, listOpts...); err != nil {
```
ASBishop, line 240: It's no longer necessary to use context.Background() whenever the context is now available as the 'ctx' input parameter. This change needs to be made in multiple locations for all controllers.

ASBishop (commented):
+1

ASBishop (approved):

PR #259, opened 2023-09-04
Improve the backend samples

Akrog:
This PR improves the Cinder backend samples to provide samples for additional storage vendors as well as make the samples require less maintenance.

In this PR we stop having an in-tree `OpenStackControlPlane` definition and instead we reference the one from the `openstack-operator` repository. This way it will always be up to date when we run `oc kustomize`.

There are 8 new backend samples from 3 vendors (HPE, Pure, NetApp) and with 4 different transport protocols (iSCSI, FC, NFS, NVMe-TCP).

### DIFF
--- a/config/samples/backends/README.md
+++ b/config/samples/backends/README.md
@@ -26,6 +26,9 @@ Currently available samples are:
 - LVM using NVMe-TCP
 - HPE 3PAR iSCSI
 - HPE 3PAR FC
+- NetApp ONTAP iSCSI
+- NetApp ONTAP FC
+- NetApp ONTAP NFS
 
 **NOTE**: These examples are designed to be applied one at time. If you attempt
 to add a second backend by applying a second example it will result in the
--- a/config/samples/backends/README.md
+++ b/config/samples/backends/README.md
@@ -24,6 +24,8 @@ Currently available samples are:
 - NFS
 - LVM using iSCSI
 - LVM using NVMe-TCP
+- HPE 3PAR iSCSI
+- HPE 3PAR FC
 
 **NOTE**: These examples are designed to be applied one at time. If you attempt
 to add a second backend by applying a second example it will result in the
@@ -62,6 +64,16 @@ we can do everything ourselves.
 If we already have a deployment working we can always use
 `oc kustomize lvm/iscsi | oc apply -f -`. from this directory to make the
 changes.
+
+Some backends, like 3PAR, require a custom container because they have external
+dependencies. So there is a need to build a specific container. In RHOSP
+vendors will provide certified containers through Red Hat's container image
+registry. For illustration and development purposes this repository provides
+samples of what a `Dockerfile` would look like for each of the vendor's. This
+`Dockerfile` will be present in the vendor's directory (e.g. `hpe/Dockerfile`),
+so we would need to build a container, make it available in a registry, and
+then provide it to the cinder operator via the `containerImage` as seen in the
+samples.
 
 ## Ceph example
 
--- a/config/samples/backends/README.md
+++ b/config/samples/backends/README.md
@@ -5,10 +5,14 @@ the `kustomize` configuration management tool available through the `oc
 kustomize` command.
 
 These samples are not meant to serve as deployment recommendations, just as
-working examples to serve as reference.
+working examples to serve as reference, and they use a base OpenStack
+configuration that is unlikely to match any real deployment.
 
 For each backend there will be a `backend.yaml` file containing an overlay for
 the `OpenStackControlPlane` with just the storage related information.
+
+They will usually also have a secrets `yaml` file where we need to replace the
+sample credentials with our real storage ones.
 
 Backend pre-requirements will be listed in that same `backend.yaml` file.
 These can range from having to replace the storage system's address and
@@ -21,28 +25,58 @@ Currently available samples are:
 - LVM using iSCSI
 - LVM using NVMe-TCP
 
+**NOTE**: These examples are designed to be applied one at time. If you attempt
+to add a second backend by applying a second example it will result in the
+first backend being removed. It is certainly possible to define a deployment
+with multiple backends, but that requires a single CR that defines all
+backends.
+
+## Using Examples
+
+Once we have `crc` running making an OpenStack deployment using one of the
+example backends is trivial and it's almost the same as any other deployment
+with the addition of creating and exporting the manifest to use:
+
+```
+$ cd install_yamls
+$ make crc_storage openstack
+$ oc kustomize ../cinder-operator/config/samples/backends/lvm/iscsi > ~/openstack-deployment.yaml
+$ export OPENSTACK_CR=`realpath ~/openstack-deployment.yaml`
+$ make openstack_deploy
+```
+
+Be aware that the some of the Cinder examples (mostly all but NFS and Ceph)
+will reboot your OpenShift cluster because they use `MachineConfig` manifests
+(to deploy things like `iscsid` and `multipathd`) that require a reboot to be
+applied due to the immutability of CoreOS (the underlying OpenShift operating
+system).  This means that the deployment takes longer and the cluster will stop
+responding for a bit.
+
+The reboot will only happen the first time the `MachineConfig` is applied.
+
+For real deployments OpenShift administration may not fall under our
+responsibilities, so we may need to reach to someone else to apply the
+`MachineConfig`, but for our samples we assume we are on a dev environment and
+we can do everything ourselves.
+
+If we already have a deployment working we can always use
+`oc kustomize lvm/iscsi | oc apply -f -`. from this directory to make the
+changes.
+
 ## Ceph example
 
-Once the OpenStack operators are running in your OpenShift cluster and
-the secret `osp-secret` is present, one can deploy OpenStack with a
-specific storage backend with single command.  For example for Ceph we can do:
-`oc kustomize ceph | oc apply -f -`.
-
-The result of the `oc kustomize ceph` command is a complete
-`OpenStackControlPlane` manifest, and we can see its contents by redirecting it
-to a file or just piping it to `less`: `oc kustomize ceph | less`.
-
-Creating the basic secret that our samples require can be done using the
-`install_yamls` target called `input`.
-
-A complete example when we already have CRC running would be:
+For Ceph we need an extra step to deploy the Ceph cluster. In our case it will
+be a toy cluster sufficient for minor PoC tests deployed with `make ceph
+TIMEOUT=90`, instead of a user-supplied external Ceph cluster. So the steps
+are:
 
 ```
 $ cd install_yamls
 $ make ceph TIMEOUT=90
-$ make crc_storage openstack input
-$ cd ../cinder-operator
-$ oc kustomize config/samples/backends/ceph | oc apply -f -
+$ make crc_storage openstack
+$ oc kustomize ../cinder-operator/config/samples/backends/lvm/iscsi > ~/openstack-deployment.yaml
+$ export OPENSTACK_CR=`realpath ~/openstack-deployment.yaml`
+$ make openstack_deploy
 ```
 
 ## Adding new samples
--- a/config/samples/backends/README.md
+++ b/config/samples/backends/README.md
@@ -29,6 +29,9 @@ Currently available samples are:
 - NetApp ONTAP iSCSI
 - NetApp ONTAP FC
 - NetApp ONTAP NFS
+- Pure Storage iSCSI
+- Pure Storage FC
+- Pure Storage NVMe-RoCE
 
 **NOTE**: These examples are designed to be applied one at time. If you attempt
 to add a second backend by applying a second example it will result in the
@@ -68,15 +71,15 @@ If we already have a deployment working we can always use
 `oc kustomize lvm/iscsi | oc apply -f -`. from this directory to make the
 changes.
 
-Some backends, like 3PAR, require a custom container because they have external
-dependencies. So there is a need to build a specific container. In RHOSP
-vendors will provide certified containers through Red Hat's container image
-registry. For illustration and development purposes this repository provides
-samples of what a `Dockerfile` would look like for each of the vendor's. This
-`Dockerfile` will be present in the vendor's directory (e.g. `hpe/Dockerfile`),
-so we would need to build a container, make it available in a registry, and
-then provide it to the cinder operator via the `containerImage` as seen in the
-samples.
+Some backends, like 3PAR and Pure, require a custom container because they have
+external dependencies. So there is a need to build a specific container. In
+RHOSP, vendors will provide certified containers through Red Hat's container
+image registry. For illustration and development purposes this repository
+provides samples of what a `Dockerfile` would look like for each of the
+vendor's. This `Dockerfile` will be present in the vendor's directory (e.g.
+`hpe/Dockerfile` and `pure/Dockerfile`), so we would need to build a container,
+make it available in a registry, and then provide it to the cinder operator via
+the `containerImage` as seen in the samples.
 
 ## Ceph example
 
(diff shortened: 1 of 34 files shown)


config/samples/backends/README.md
```diff
@@ -20,29 +20,54 @@ Currently available samples are:
...
 - LVM using iSCSI
 - LVM using NVMe-TCP
+- HPE 3PAR iSCSI
+- HPE 3PAR FC
+- NetApp ONTAP iSCSI
+- NetApp ONTAP FC
+- NetApp ONTAP NFS
+- Pure Storage iSCSI
+- Pure Storage FC
+- Pure Storage NVMe-TCP
```
sdodsley, line 30: TCP is not availble until Bobcat. You can use NVMe-RoCE instead which is supported in Antelope

config/samples/backends/pure/nvme-tcp/backend.yaml
```diff
@@ -0,0 +1,32 @@
+# To be able to use this sample it is necessary to:
+# - Have a Pure backend with iSCSI support
```
sdodsley, line 2: NVMe-RoCE, not iSCSI

config/samples/backends/pure/nvme-tcp/backend.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - Have the Pure storage credentials in cinder-volume-pure-secrets.yaml
+# - Having iscsid and multpathd running on the host (done by this sample)
```
sdodsley, line 4: iscsid not required for NVMe

config/samples/backends/pure/nvme-tcp/backend.yaml
```diff
@@ -0,0 +1,32 @@
...
+          containerImage: quay.io/geguileo/podified-antelope-openstack-cinder-volume-pure:current-podified
+          # Configuration details available at
+          # https://pure-storage-openstack-docs.readthedocs.io/en/latest/cinder/ch_cinder-configuration.html
+          customServiceConfigSecrets:
+            - cinder-volume-pure-secrets
+          customServiceConfig: |
+            [pure]
+            volume_backend_name=pure
+            volume_driver=cinder.volume.drivers.pure.PureNVMEDriver
+            pure_nvme_transport=tcp
```
sdodsley, line 32: pure_nvme_transport=roce

sdodsley (changes requested):
You will need to rename the nvme-tcp directory to nvme-roce

config/samples/backends/README.md
Akrog: Thanks! I incorrectly used the doc reference from master.

config/samples/backends/pure/nvme-tcp/backend.yaml
Akrog: True, it requires the `nvme-fabrics` kernel module

config/samples/backends/pure/nvme-roce/backend.yaml
```diff
@@ -0,0 +1,32 @@
+# To be able to use this sample it is necessary to:
+# - Have a Pure backend with NVMe-RoCE support
+# - Have the Pure storage credentials in cinder-volume-pure-secrets.yaml
+# - Having nvme-fabrics kernel module loaded on the host (done by this sample)
```
sdodsley, line 4: You should add into the docs about the nvme-rdma module that is needed along with nvme-fabrics? Does the `nvme-fabrics.yaml` you mention have nvme-fdma in as well? Given there are other drivers that need the `nvme-tcp` module as well, it might be adding all 3 kernel modules by default

sdodsley (changes requested):

config/samples/backends/pure/nvme-roce/backend.yaml
Akrog: We have a `MachineConfig` that ensures that `nvme-fabrics` is loaded ([link](https://github.com/openstack-k8s-operators/cinder-operator/blob/main/config/samples/backends/bases/nvmeof/nvme-fabrics.yaml))
That file is added automatically because we have the line `  - ../../bases/nvmeof` in this Pure NVMe-RoCE backend's `kustomization.yaml` ([link](https://github.com/openstack-k8s-operators/cinder-operator/blob/8ccadb31ea95f030c1fae434db0145cd7c379dcd/config/samples/backends/pure/nvme-roce/kustomization.yaml#L2))
We don't need to manually load the kernel modules for the individual NVMe transport protocols, they are automatically loaded on usage by the `nvme-fabrics` one.  I tested this with the LVM nvme-tcp example ([link](https://github.com/openstack-k8s-operators/cinder-operator/tree/main/config/samples/backends/lvm/nvme-tcp)).
I could be wrong and it only works when using `nvmet` (like we do in LVM) but I don't think that's the case.

config/samples/backends/bases/openstack/kustomization.yaml
```diff
@@ -1,2 +1,24 @@
 resources:
-  - openstack.yaml
+  - https://raw.githubusercontent.com/openstack-k8s-operators/openstack-operator/main/config/samples/core_v1beta1_openstackcontrolplane_galera_network_isolation.yaml
```
ASBishop, line 2: Ooh, this is good! I like the way this references the openstack-operator's sample, rather than the original hard coded one in the cinder-operator which is likely to drift out of sync with the rest of the operators.

config/samples/backends/ceph/backend.yaml
```diff
@@ -29,6 +29,8 @@ spec:
     template:
       cinderVolumes:
         ceph:
+          networkAttachments:
+          - storage
           # Once nova is enabled we need to either manually define the
           # rbd_secret_uuid configuration option or use Antelope release.
```
ASBishop, line 35: Can this comment be eliminated now that we're using Antelope containers>

config/samples/backends/hpe/3par/fc/cinder-volume-3par-secrets.yaml
```diff
@@ -0,0 +1,22 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  labels:
+    service: cinder
+    component: cinder-volume
+  name: cinder-volume-3par-secrets
+type: Opaque
+stringData:
+  hpe3par-secrets.conf: |
```
ASBishop, line 12: Were you going to sanitize the secrets files? The presumption is every entry (L14..L22) by definition needs to be filled in by the user.

config/samples/backends/hpe/3par/fc/backend.yaml
```diff
@@ -0,0 +1,33 @@
...
+    template:
+      cinderVolumes:
+        hpe3par-fc:
+          networkAttachments:
+          - storage
+          containerImage: quay.io/geguileo/podified-antelope-openstack-cinder-volume-3par:current-podified
+          customServiceConfigSecrets:
+            - cinder-volume-3par-secrets
+          customServiceConfig: |
+            [hpe3par]
```
ASBishop, line 27: Given that we're offering both FC and iscsi examples, should the backend section (and perhaps the volume_backend_name) distinguish between them, as you do in the CR name given on L20?

That reminds me: the documentation should emphasize the examples are designed to be applied one at a time, and you can't get two backends by applying a second one. Each sample you apply will cause an existing c-vol backend to be removed.

config/samples/backends/netapp/ontap/nfs/backend.yaml
```diff
@@ -0,0 +1,50 @@
...
+# - Have a NetApp ONTAP backend with NFS support
+# - Have the NetApp storage credentials and NFS configuration in cinder-volume-netapp-secrets.yaml
+
+apiVersion: core.openstack.org/v1beta1
+kind: OpenStackControlPlane
+metadata:
+  name: openstack
+spec:
+  # We need to "drop" the "nfs_shares_config" file into the volume service.
+  # The contents come from a secret that we are creting in the
```
ASBishop, line 11: typo: "creting" should be "creating"

config/samples/backends/netapp/ontap/nfs/backend.yaml
```diff
@@ -0,0 +1,50 @@
...
+            - cinder-volume-ontap-secrets
+          customServiceConfig: |
+            [ontap]
+            volume_backend_name=ontap
+            volume_driver=cinder.volume.drivers.netapp.common.NetAppDriver
+            netapp_server_hostname=hostname
+            netapp_server_port=80
+            netapp_storage_protocol=nfs
+            netapp_storage_family=ontap_cluster
+            nfs_shares_config=/etc/cinder/nfs_shares.d/shares-config
```
ASBishop, line 50: I'm sure this is fine, but fwiw it's different than how tripleo handles the shares file. TripleO would deploy a file named /etc/cinder/shares.conf

config/samples/backends/pure/fc/backend.yaml
```diff
@@ -0,0 +1,31 @@
+# To be able to use this sample it is necessary to:
+# - Have a Pure backend with iSCSI support
+# - Have the Pure storage credentials in cinder-volume-pure-secrets.yaml
+# - Having iscsid and multpathd running on the host (done by this sample)
```
ASBishop, line 4: iscsid, in an FC example?

config/samples/backends/hpe/3par/fc/backend.yaml
```diff
@@ -0,0 +1,33 @@
...
+metadata:
+  name: openstack
+spec:
+  cinder:
+    template:
+      cinderVolumes:
+        hpe3par-fc:
+          networkAttachments:
+          - storage
+          containerImage: quay.io/geguileo/podified-antelope-openstack-cinder-volume-3par:current-podified
```
ASBishop, line 23: We probably need to provide instructions (or links to them) for building a pushing these custom c-vol images. Users will need to update L23 with the URL to the image they built and pushed.

config/samples/backends/hpe/3par/iscsi/backend.yaml
```diff
@@ -0,0 +1,33 @@
...
+metadata:
+  name: openstack
+spec:
+  cinder:
+    template:
+      cinderVolumes:
+        hpe3par-iscsi:
+          networkAttachments:
+          - storage
+          containerImage: quay.io/geguileo/podified-antelope-openstack-cinder-volume-3par:current-podified
```
ASBishop, line 23: Here's another "geguileo" reference you may want to revise :-/

config/samples/backends/pure/fc/backend.yaml
```diff
@@ -0,0 +1,31 @@
...
+metadata:
+  name: openstack
+spec:
+  cinder:
+    template:
+      cinderVolumes:
+        pure-fc:
+          networkAttachments:
+          - storage
+          containerImage: quay.io/geguileo/podified-antelope-openstack-cinder-volume-pure:current-podified
```
ASBishop, line 23: Another "geguileo"

config/samples/backends/pure/iscsi/backend.yaml
```diff
@@ -0,0 +1,31 @@
...
+metadata:
+  name: openstack
+spec:
+  cinder:
+    template:
+      cinderVolumes:
+        pure-iscsi:
+          networkAttachments:
+          - storage
+          containerImage: quay.io/geguileo/podified-antelope-openstack-cinder-volume-pure:current-podified
```
ASBishop, line 23: Another "geguileo"

config/samples/backends/pure/nvme-roce/backend.yaml
```diff
@@ -0,0 +1,32 @@
...
+metadata:
+  name: openstack
+spec:
+  cinder:
+    template:
+      cinderVolumes:
+        pure-nvme-roce:
+          networkAttachments:
+          - storage
+          containerImage: quay.io/geguileo/podified-antelope-openstack-cinder-volume-pure:current-podified
```
ASBishop, line 23: Another one!

config/samples/backends/bases/openstack/kustomization.yaml
Akrog: And this gets even better when you look at the glance using cinder samples I wrote, because they can reference the whole cinder LVM iscsi sample and only change the glance part: https://github.com/openstack-k8s-operators/glance-operator/blob/main/config/samples/backends/cinder/lvm-iscsi/kustomization.yaml

config/samples/backends/hpe/3par/fc/cinder-volume-3par-secrets.yaml
Akrog: Yes, they should be filled in by the user, but I'm using the samples from the vendor's documentation.

config/samples/backends/hpe/3par/fc/backend.yaml
Akrog: I didn't provide step by step instructions, but I did mention it in the `README.md` file, though on second look it was too brief. I'll add a couple more words there.

config/samples/backends/hpe/3par/iscsi/backend.yaml
Akrog: I did that on purpose, because those containers actually exist and can be used by people trying those samples. ;-)

config/samples/backends/pure/nvme-roce/backend.yaml
Akrog: Just to be in the safe side I have added a patch to the PR that loads the other nvme kernel modules supported by openstack: https://github.com/openstack-k8s-operators/cinder-operator/pull/259/commits/c076183ffae963f10bd1c2d722842a344a7c0406

config/samples/backends/hpe/3par/iscsi/backend.yaml
Akrog: Yes, I think it's worth putting a comment over it saying that that specific image was built using the sample `Dockerfile`

config/samples/backends/netapp/ontap/nfs/backend.yaml
Akrog: Done

config/samples/backends/netapp/ontap/nfs/backend.yaml
Akrog: I'm doing it this way based on how I'm mounting the secret on L32.  I could change it to do it how you describe it, but then this YAML would need a bit more detail on the mount side, and I'd rather keep it as simple as possible.

config/samples/backends/hpe/3par/fc/backend.yaml
Akrog: I didn't distinguish it for any of the other vendor samples either, and I think it's not worth it.  In my own deployments I don't usually distinguish them when I only have 1 transport protocol from the array.

I have added a new paragraph on the README.md file about the single sample thingy.

config/samples/backends/README.md
```diff
@@ -20,29 +24,74 @@ Currently available samples are:
...
+- HPE 3PAR FC
+- NetApp ONTAP iSCSI
+- NetApp ONTAP FC
+- NetApp ONTAP NFS
+- Pure Storage iSCSI
+- Pure Storage FC
+- Pure Storage NVMe-RoCE
 
-## Ceph example
+These examples are designed to be applied one at a time, you cannot get 2
```
Akrog, line 36: @ASBishop here is where I've added the comment of this being single backend samples.

ASBishop: I'd like to see something more like this (for your consideration):

NOTE: These examples are designed to be applied one at time. If you attempt to add a second backend by applying a second example, it will result in the first backend being removed. It is certainly possible to define a deployment with multiple backends, but that requires a single CR that defines all backends.

Akrog: Done

ASBishop (approved):
/lgtm

PR #266, opened 2023-09-15
enable service token role validation for all service tokens

SeanMooney:
As part of adressing CVE-2023-2088 cinder was modifed to
require the service role to be present in service token when
calling the attachemtn api to modify attachments related to nova
instance. One recomendation of that CVE mitigration discussions
was that all services shoudl enabel the service token role validation
by default. This change simplely enabled that by setting
[keystone_authtoken]/service_token_roles_required = true

Related: OSPRH191

### DIFF
--- a/templates/cinderapi/config/01-service-defaults.conf
+++ b/templates/cinderapi/config/01-service-defaults.conf
@@ -1,2 +1,9 @@
 [DEFAULT]
 log_file = {{ .LogFile }}
+
+[keystone_authtoken]
+# This is part of hardening related to CVE-2023-2088
+# when enabled the service token user must have the service role to be considered valid.
+# cinder already checks for this, explicitly in the case of the attchment API even when
+# this is not enforced for all service token validation.
+service_token_roles_required = true


templates/cinderapi/config/01-service-defaults.conf
```diff
@@ -1,2 +1,9 @@
 [DEFAULT]
 log_file = {{ .LogFile }}
+
+[keystone_authtoken]
+# This is part of hardening related to CVE-2023-2088
+# when enabled the service token user must have the service role to be considered valid.
+# cinder already checks for this, explicitly in the case of the attchment API even when
+# this is not enforced for all service token validation.
+service_token_roles_required = true
```
ASBishop, line 9: Looks like this needs a newline.

ASBishop (changes requested):

templates/cinderapi/config/01-service-defaults.conf
SeanMooney: i guess i did not activate the venv when committing. i have precommet but it only runs if i have the venv activated
ill respin this shortly

Akrog: We already added this in [PR#257](https://github.com/openstack-k8s-operators/cinder-operator/pull/257/files) and it can be seen here (without the CVE related comment): https://github.com/openstack-k8s-operators/cinder-operator/blob/e8e5fb3c23c3a6271d48db593ad89e2f9ea43c1a/templates/cinder/config/00-global-defaults.conf#L66

If I remember correctly the cinder CVE code doesn't really need that option, it has an additional internal check so it wouldn't have to depend on this config option. I enabled this option in my PR anyway because I like the keystone middleware being the one that gives the error..

Akrog (changes requested):
Thank you for keeping an eye on this, though we already have it. :-)

templates/cinderapi/config/01-service-defaults.conf
SeanMooney: oh cool in that case ill abandon this pr

SeanMooney:
fixed by #257

PR #276, opened 2023-10-03
Fix NetworkAttachments on Cinder Volume

Akrog:
When we deploy multiple cinder volume backends with network attachments the reconciliation loop is constantly getting called.

This issue happens because each of the cindervolume objects ends up with the IP addresses of the network attachments of all the different backends as the labels that these services are the same.

This patch adds a new label with the backend name so that the network attachments are not merged.

Before the patch we would see both IP Addresses:

```
NAME                         NETWORKATTACHMENTS                                                                                  STATUS   MESSAGE
cinder-volume-lvm-iscsi      {"openshift-sdn":["10.217.1.74","10.217.1.72"],"openstack/storage":["172.18.0.32","172.18.0.30"]}   True     Setup complete
cinder-volume-lvm-nvme-tcp   {"openshift-sdn":["10.217.1.74","10.217.1.76"],"openstack/storage":["172.18.0.32","172.18.0.30"]}   True     Setup complete
```

After the patch we would see:

```
NAME                         NETWORKATTACHMENTS                                                       STATUS   MESSAGE
cinder-volume-lvm-iscsi      {"openshift-sdn":["10.217.0.211"],"openstack/storage":["172.18.0.32"]}   True     Setup complete
cinder-volume-lvm-nvme-tcp   {"openshift-sdn":["10.217.0.210"],"openstack/storage":["172.18.0.30"]}   True     Setup complete
```

### DIFF
--- a/controllers/cindervolume_controller.go
+++ b/controllers/cindervolume_controller.go
@@ -310,6 +310,7 @@ func (r *CinderVolumeReconciler) reconcileNormal(ctx context.Context, instance *
 	serviceLabels := map[string]string{
 		common.AppSelector:       cinder.ServiceName,
 		common.ComponentSelector: cindervolume.Component,
+		cindervolume.Backend:     instance.Name[len(cindervolume.Component)+1:],
 	}
 
 	//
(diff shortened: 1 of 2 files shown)


ASBishop (approved):
/lgtm

PR #279, opened 2023-10-09
Configure target IP addresses for LVM backend

ASBishop:
Configure the cinder-volume service's target_ip_address and target_secondary_ip_addresses using addresses in the network attachment status. This is necessary in order for EDPM compute nodes to be able to make volume attachment requests (via iSCSI or NVMe/TCP) over the storage network.

Jira: OSP-28445
(cherry picked from commit 7cd5920d8f959aa86d85b8ca37c369239c0d09e7)


ASBishop:
Missed the dev preview 2 deadline.

PR #286, opened 2023-10-25
Enable caching using memcached

ASBishop:
Jira: [OSP-26501](https://issues.redhat.com//browse/OSP-26501)

### DIFF
--- a/controllers/cinder_controller.go
+++ b/controllers/cinder_controller.go
@@ -19,10 +19,12 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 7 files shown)


ASBishop:
/hold

I need to verify something

controllers/cinder_controller.go
```diff
@@ -780,6 +851,7 @@ func (r *CinderReconciler) generateServiceConfigs(
 	instance *cinderv1beta1.Cinder,
 	envVars *map[string]env.Setter,
 	serviceLabels map[string]string,
+	mc *memcachedv1.Memcached,
```
Akrog, line 854: nit: Wouldn't it make it more readable using `memcached` instead of `mc`?

Akrog (changes requested):
The code looks good, but it is making the deployment of memcached a hard requirement.
I would prefer an approach that would make it optional, for example when the `memcachedInstance` is set to an empty string the `enabled` in the global-defaults is set to `false`.

ASBishop (commented):
While I don't know exactly where the discussion took place, my understanding is memcached is a required component (just as it is in tripleo). I recall seeing (but cannot find) a comment in another operator's PR that it is not necessary, at least at this juncture, to provide support for conditionally configuring memcached. See also https://issues.redhat.com/browse/OSP-26273

ASBishop:
/retest

'[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1129)'

Akrog (approved):
/lgtm

PR #295, opened 2023-11-13
Add cronjob to periodically purge the DB

ASBishop:
The cron schedule and the age at which records are purged are configurable in the CRD, with the same default values chosen for TripleO based deployments.


fmount (approved):
/lgtm

pkg/cinderscheduler/statefulset.go
```diff
@@ -35,16 +35,16 @@ const (
...
 func StatefulSet(
-	instance *cinderv1beta1.CinderScheduler,
+	instance *cinderv1.CinderScheduler,
 	configHash string,
 	labels map[string]string,
 	annotations map[string]string,
 ) *appsv1.StatefulSet {
 	rootUser := int64(0)
 	// Cinder's uid and gid magic numbers come from the 'cinder-user' in
 	// https://github.com/openstack/kolla/blob/master/kolla/common/users.py
```
Akrog, line 45: -1: Remove these 2 comment lines

pkg/cinderbackup/statefulset.go
```diff
@@ -44,8 +44,8 @@ func StatefulSet(
 	rootUser := int64(0)
 	// Cinder's uid and gid magic numbers come from the 'cinder-user' in
 	// https://github.com/openstack/kolla/blob/master/kolla/common/users.py
```
Akrog, line 46: -1: Remove this from here.
nit: Move it to the constant definition as mentioned above.

pkg/cinder/cronjob.go
```diff
@@ -0,0 +1,131 @@
...
+	labels map[string]string,
+	annotations map[string]string,
+) *batchv1.CronJob {
+	cinderUser := int64(cinderv1.CinderUserID)
+	cinderGroup := int64(cinderv1.CinderGroupID)
+	config0644AccessMode := int32(0644)
+
+	debugArg := ""
+	if instance.Spec.Debug.DBPurge {
+		debugArg = " --debug"
```
Akrog, line 40: ?: I thought that the `debug` section elements was meant to pause the execution to let the human operator/developer jump into the container and debug things manually.  Are we changing its meaning now?

pkg/cinder/cronjob.go
```diff
@@ -0,0 +1,131 @@
...
+			Annotations: annotations,
+			Labels:      labels,
+		},
+		Spec: batchv1.CronJobSpec{
+			Schedule:          instance.Spec.DBPurge.Schedule,
+			ConcurrencyPolicy: batchv1.ForbidConcurrent,
+			JobTemplate: batchv1.JobTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: annotations,
+					Labels:      labels,
```
Akrog, line 92: Thank you for ensuring everything gets the labels.

controllers/cinder_controller.go
```diff
@@ -819,6 +823,27 @@ func (r *CinderReconciler) reconcileNormal(ctx context.Context, instance *cinder
 		return ctrl.Result{}, err
 	}
 
+	// create CronJob
+	cronjobDef := cinder.CronJob(instance, serviceLabels, serviceAnnotations)
+	cronjob := cronjob.NewCronJob(
+		cronjobDef,
+		5*time.Second,
```
Akrog, line 830: nit: Since this is not a job we need "right now" we can probably set a higher reconciliation timeout value.

Akrog (changes requested):
Thanks for this nice feature!
I have added some questions and change requests.

api/v1beta1/cinder_types.go
```diff
@@ -18,12 +18,16 @@ package v1beta1
 
 import (
 	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
-	"github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	"github.com/openstack-k8s-operators/lib-common/modules/storage"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
+	// CinderUserID - Kolla's cinder UID
```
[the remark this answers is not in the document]
ASBishop: Done

api/v1beta1/cinder_types.go
```diff
@@ -172,6 +185,32 @@ type Cinder struct {
 	Status CinderStatus `json:"status,omitempty"`
 }
 
+// DBPurge struct is used to model the parameters exposed to the Manila API CronJob
```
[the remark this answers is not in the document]
ASBishop: Done

api/v1beta1/cinder_webhook.go
```diff
@@ -86,6 +97,11 @@ func (spec *CinderSpec) Default() {
 		// This is required, as the loop variable is a by-value copy
 		spec.CinderVolumes[index] = cinderVolume
 	}
+
+	if (spec.DBPurge.Age == 0) && (spec.DBPurge.Schedule == "") {
```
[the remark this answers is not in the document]
ASBishop: I'm not sure if it's possible, but I see no reason why these shouldn't be independently checked, so I'll update the code.

controllers/cinder_controller.go
ASBishop: Sure, but what would that value be?

pkg/cinder/cronjob.go
ASBishop: Well, I did update the comment (see cinder_types.go:203). I know the original use case was to pause execution, but "debug" is a generic term that can be broadly applied. I don't think it would be fair for the first use case (pausing execution) to own the rights to the entire ```debug``` section.

pkg/cinderscheduler/statefulset.go
ASBishop: Done

pkg/cindervolume/statefulset.go
```diff
@@ -44,8 +44,8 @@ func StatefulSet(
 	rootUser := int64(0)
 	// Cinder's uid and gid magic numbers come from the 'cinder-user' in
 	// https://github.com/openstack/kolla/blob/master/kolla/common/users.py
```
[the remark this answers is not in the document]
ASBishop: Done

ASBishop:
/retest

looks like temporary issues with ci infrastructure

api/v1beta1/cinder_webhook.go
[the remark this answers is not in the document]
fmount: yeah, I think the main problem we were trying to solve here is when the CR is applied and no `dbPurge` related section is added. `kubebuilder` is basically not initializing the fields of the structs with the default values, and itresulted in something like `dbPurge: {}`.
However, without digging more into the original issue, I think that the latest @ASBishop's PS is good enough, and processing the two params independently works fine (and covers the previous case I mentioned).
I'm going to update the `Manila` patch the same way, thanks for the great input!

pkg/cinder/cronjob.go
Akrog: That also makes sense, but would getting the debug info in the cron job be really helpful in the field instead of being able to get into the container from the cron job itself and debuging things manually?

ASBishop: I believe the intent is to get the db_purge command to generate log messages. If they need to actually debug what happens when running the command then you don't actually need to pause the container to do that. You can rsh into a running container and manually run the command there.

controllers/cinder_controller.go
Akrog: Good question, I don't know :-P

api/v1beta1/cinder_types.go
```diff
@@ -172,6 +186,32 @@ type Cinder struct {
...
+	// Age is the DBPurgeAge parameter and indicates the number of days of purging DB records
+	Age int `json:"age"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="1 0 * * *"
+	// Schedule defines the crontab format string to schedule the DBPurge cronJob
+	Schedule string `json:"schedule"`
+}
+
+// CinderDebug indicates whether certain stages of Cinder deployment should
+// pause in debug mode, or if debug is enabled when purging the DB.
```
Akrog, line 203: -1: If we are no longer going to pause on these actions we should rephrase this.

api/v1beta1/cinder_types.go
```diff
@@ -172,6 +186,32 @@ type Cinder struct {
...
+type CinderDebug struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// dbSync enable debug
```
Akrog, line 207: Mention that this debug action pauses the container.

api/v1beta1/cinder_types.go
```diff
@@ -172,6 +186,32 @@ type Cinder struct {
...
+	DBSync bool `json:"dbSync"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// DBPurge enable debug
+	DBPurge bool `json:"dbPurge"`
```
Akrog, line 212: Mention that this debug action just increases the verbosity of the command executed.

pkg/cinder/cronjob.go
Akrog: I'm not entirely convinced there's no value in pausing the cron job, but I don't have a strong opinion, so this code can stay as it is.

Since we are changing the pattern of pausing the container I think we should follow @fmount's suggestion and rephrase the comments to mention what kind of debugging each flag achieves.

api/v1beta1/cinder_types.go
ASBishop: Um, I thought I did. But I can remove this comment in favor of explicitly describing each flag within the structure.

api/v1beta1/cinder_types.go
ASBishop: Done

api/v1beta1/cinder_types.go
ASBishop: Done

api/v1beta1/cinder_types.go
```diff
@@ -172,6 +186,32 @@ type Cinder struct {
...
+	// Schedule defines the crontab format string to schedule the DBPurge cronJob
+	Schedule string `json:"schedule"`
+}
+
+// CinderDebug contains flags related to multiple debug activities. See the
+// individual comments for what this means for each flag.
+type CinderDebug struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	// DBSync pauses the dbSync container prior to executing the db_sync command.
```
Akrog, line 207: -1: I hate to downvote for this, but the comment is not correct, [it's an infinite sleep](https://github.com/openstack-k8s-operators/lib-common/blob/6bd6025ade373131acb7737c58cb61b43dafeee4/modules/common/const.go#L32), so it will never execute the dbsync command.
Maybe it's not the best command, and we should have a command to sleep until a file is removed, like I did in [openstack-must-gather](https://github.com/openstack-k8s-operators/openstack-must-gather/blob/main/collection-scripts/gather_debug), but it's what we have right now.

Akrog (changes requested):

api/v1beta1/cinder_types.go
ASBishop: Changing "prior to executing" to "instead of executing" per our conversation.

Akrog (approved):
/lgtm

PR #305, opened 2023-12-06
Configure barbican access

ASBishop:
Add the [barbican] settings so that all cinder services can access barbican via its internal endpoint. Barbican is already the default key manager.

Jira: [OSPRH-1376](https://issues.redhat.com//browse/OSPRH-1376)


fmount (approved):
Looks good to me, and it seems aligned with what we've done in both [Glance PR](https://github.com/openstack-k8s-operators/glance-operator/pull/378) and [Nova](https://github.com/openstack-k8s-operators/nova-operator/pull/609).

PR #309, opened 2023-12-13
Switch cinder-api to a statefulset

ASBishop:
Change cinder-api from a deployment to a statefulset. Although the service doesn't require preserving stateful data, it helps with the pod naming convention so they're similar to glance-api pods, and eliminates the squirrely suffixes like "cinder-api-xzkpl"

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -30,18 +30,6 @@ rules:
 - apiGroups:
   - apps
   resources:
-  - deployments
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - apps
-  resources:
   - statefulsets
   verbs:
   - create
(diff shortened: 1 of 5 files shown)


api/bases/cinder.openstack.org_cinderapis.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.11.1
+    controller-gen.kubebuilder.io/version: v0.10.0
```
fmount, line 6: you probably have an older version of `controller-gen` within your repo because this shouldn't change. 
(the same comment applies to the other files).

fmount (changes requested):
In general I'm +1 to move to `statefulset` and be consistent across the board. I don't see a particular reason that prevents this kind of switch.
We'll make another pass once the issues  w/ controller-gen are resolved, but I'm not against this change.

api/bases/cinder.openstack.org_cinderapis.yaml
ASBishop: whoa! I knew I was working on an older system that has an old controller-gen, but accidentally "git add -u" when I only meant to update a single file.

fmount (approved):
/lgtm

PR #311, opened 2023-12-14
Fix our backend samples

Akrog:
Our backend samples no longer work on OCP 4.14 and they fail the `oc kustomize` call with:

 error: invalid Kustomization: json: cannot unmarshal string into Go
 struct field Kustomization.patches of type types.Patch

In this patch we update the `patches` section from:
```
 patches:
   - backend
```
to
```
 patches:
   - path: backend
```


ASBishop (approved):
/lgtm

PR #317, opened 2024-01-04
Unpin module golang.org/x/net

ASBishop:
It was previously bumped to v0.18.0 in order to avoid the Rapid Reset CVE, but we're past that now and the latest version is v0.19.0.


Akrog (approved):
/lgtm

dprince (approved):

PR #318, opened 2024-01-12
Enable SRBAC by default

Akrog:
This patch enables SRBAC by default instead of requiring explicit enabling like we did in OSP17.

Only the API service will include the configuration since policies are to be checked at the endpoint in Cinder API.

Jira: [OSPRH-3281](https://issues.redhat.com//browse/OSPRH-3281)


Akrog:
Holding the PR until I get confirmation that this change is sufficient.

ASBishop (approved):
/lgtm

PR #327, opened 2024-02-02
Drop Debug fields

gibizer:
These fields were redundant. Crashing pods can be debugged via oc debug.

Implements: [OSPRH-4290](https://issues.redhat.com//browse/OSPRH-4290)


Akrog:
Thanks for the PR, I'm putting it on hold until I can confirm that I can do everything I need for customer engagements with the `oc debug` feature.

gibizer:
> Thanks for the PR, I'm putting it on hold until I can confirm that I can do everything I need for customer engagements with the `oc debug` feature.

let me know if something is not possible

pkg/cinderapi/statefuleset.go
```diff
@@ -56,30 +55,20 @@ func StatefulSet(
...
-		//
-		// https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
-		//
-		livenessProbe.HTTPGet = &corev1.HTTPGetAction{
-			Path: "/healthcheck",
-			Port: intstr.IntOrString{Type: intstr.Int, IntVal: int32(cinder.CinderPublicPort)},
-		}
-		readinessProbe.HTTPGet = livenessProbe.HTTPGet
+	args := []string{"-c", ServiceCommand}
+	args = append(args, ServiceCommand)
```
gibizer, line 59: I need to drop this

pkg/cinderapi/statefuleset.go
```diff
@@ -113,6 +101,7 @@ func StatefulSet(
 			volumeMounts = append(volumeMounts, svc.CreateVolumeMounts(endpt.String())...)
 		}
 	}
+	readinessProbe.HTTPGet = livenessProbe.HTTPGet
```
Akrog, line 104: -1: This line is unnecessary, we already set it on L66

Akrog (changes requested):
The downvote is because there is a duplicated line in the code that should be removed.  The PR looks good to me otherwise.

I have looked into the replacement of the `debug` section with the `oc debug` command and all the primary cases seem to be covered.

I have proposed a small improvement on the documentation to explicitly show how we can do some of the debugging scenarios: https://github.com/openstack-k8s-operators/docs/pull/96

One scenario that would not be covered is debugging the db sync job pod when it doesn't fail.  If it fails we can debug it, but if the job completes and disappears then we cannot debug it.  I don't think it's a big problem, since the common case would be to debug it on failure.

pkg/cinderapi/statefuleset.go
gibizer: fixed

gibizer:
/test cinder-operator-build-deploy-kuttl
```
    logger.go:42: 12:31:26 | cinder_tls/1-deploy-cinder | Error from server (NotFound): keystoneendpoints.keystone.openstack.org "cinderv3" not found
```

gibizer:
/test cinder-operator-build-deploy-tempest
```
This file contained potentially sensitive information and has been removed.
```

Akrog (approved):
/lgtm

PR #328, opened 2024-02-02
Always use --debug for db purge

gibizer:


Akrog:
We already dropped the debug struct, so we should be able to drop this PR.
If I'm wrong, please feel free to re-open it.

gibizer:
> We already dropped the debug struct, so we should be able to drop this PR. If I'm wrong, please feel free to re-open it.

I totally forgot this. After rebase I still see meaningful content in this PR.

ASBishop:
> > We already dropped the debug struct, so we should be able to drop this PR. If I'm wrong, please feel free to re-open it.
> 
> I totally forgot this. After rebase I still see meaningful content in this PR. So I reopened it.

I agree this change makes sense.

ASBishop (approved):
/lgtm

PR #333, opened 2024-02-12
Backend Samples: place on worker nodes

Akrog:
The current samples are deploying MachineConfigs on master nodes, which works just fine because right now all our master nodes are also worker nodes, but is not technically correct.

This patch updates the MachineConfigs to deploy on worker nodes, which is the technically correct location.

### DIFF
--- a/config/samples/backends/bases/lvm/lvm.yaml
+++ b/config/samples/backends/bases/lvm/lvm.yaml
@@ -2,10 +2,10 @@ apiVersion: machineconfiguration.openshift.io/v1
 kind: MachineConfig
 metadata:
   labels:
-    machineconfiguration.openshift.io/role: master
+    machineconfiguration.openshift.io/role: worker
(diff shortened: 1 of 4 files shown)


config/samples/backends/bases/lvm/lvm.yaml
```diff
@@ -2,10 +2,10 @@ apiVersion: machineconfiguration.openshift.io/v1
 kind: MachineConfig
 metadata:
   labels:
-    machineconfiguration.openshift.io/role: master
+    machineconfiguration.openshift.io/role: worker
     service: cinder
     component: cinder-volume
```
ASBishop, line 7: Independent of this PR (which looks fine), we'll need to ensure this machineConfig is applied to only one worker, and the associated c-vol service is constrained to that worker.

ASBishop (approved):
/lgtm

PR #336, opened 2024-02-13
Add CinderSpecCore struct

dprince:
This version of the struct (called "core") is meant to be used via the OpenStackControlplane. It is the same as CinderSpec only it is missing the containerImages parameters from the nested API, Backup, Scheduler, and Volume specs.

This is useful to prevent users from interacting with these
parameters directly via the OpenStackControlplane


ASBishop (commented):
Dan, I assume this is all part of the migration to using "openstackversions" to manage container images. What's the state of that effort? We don't want to lose the ability to manage cinder-volume's container images if there will be a time lag between merging this PR and when "openstackversions" will be ready.

api/v1beta1/cinderapi_types.go
```diff
@@ -23,8 +23,36 @@ import (
...
+
 // CinderAPITemplate defines the input parameters for the Cinder API service
 type CinderAPITemplate struct {
+
+    // +kubebuilder:validation:Required
+    // ContainerImage - Cinder Container Image URL (will be set to environmental default if empty)
+    ContainerImage string `json:"containerImage"`
+
 	// Common input parameters for the Cinder API service
 	CinderServiceTemplate `json:",inline"`
```
Akrog, line 57: Why don't we use `CinderAPITemplateCore` instead?

Because L29-L45 are the same as L58-L72.

api/v1beta1/cinderbackup_types.go
```diff
@@ -22,8 +22,26 @@ import (
...
+
 // CinderBackupTemplate defines the input parameters for the Cinder Backup service
 type CinderBackupTemplate struct {
+
+    // +kubebuilder:validation:Required
+    // ContainerImage - Cinder Container Image URL (will be set to environmental default if empty)
+    ContainerImage string `json:"containerImage"`
+
 	// Common input parameters for the Cinder Backup service
 	CinderServiceTemplate `json:",inline"`
```
Akrog, line 46: Same here, we should use `CinderBackupTemplateCore` instead.

api/v1beta1/cinderscheduler_types.go
```diff
@@ -22,8 +22,25 @@ import (
...
+
 // CinderSchedulerTemplate defines the input parameters for the Cinder Scheduler service
 type CinderSchedulerTemplate struct {
+
+    // +kubebuilder:validation:Required
+    // ContainerImage - Cinder Container Image URL (will be set to environmental default if empty)
+    ContainerImage string `json:"containerImage"`
+
 	// Common input parameters for the Cinder Scheduler service
 	CinderServiceTemplate `json:",inline"`
```
Akrog, line 45: Replace with `CinderSchedulerTempalteCore`.

api/v1beta1/cindervolume_types.go
```diff
@@ -22,8 +22,27 @@ import (
...
+
 // CinderVolumeTemplate defines the input parameters for the Cinder Volume service
 type CinderVolumeTemplate struct {
+
+    // +kubebuilder:validation:Required
+    // ContainerImage - Cinder Container Image URL (will be set to environmental default if empty)
+    ContainerImage string `json:"containerImage"`
+
 	// Common input parameters for the Cinder Volume service
 	CinderServiceTemplate `json:",inline"`
```
Akrog, line 47: Replace with `CidnerVolumeeTemplateCore`

Akrog (changes requested):
Approach looks good to me.
Besides the changes requested by Alan around indentation, I think we should change CinderAPITemplate and other templates to use their Core equivalent to remove code duplication.

api/v1beta1/cinderapi_types.go
```diff
@@ -23,8 +23,36 @@ import (
...
+	TLS tls.API `json:"tls,omitempty"`
+}
+
+
 // CinderAPITemplate defines the input parameters for the Cinder API service
 type CinderAPITemplate struct {
+
+    // +kubebuilder:validation:Required
+    // ContainerImage - Cinder Container Image URL (will be set to environmental default if empty)
+    ContainerImage string `json:"containerImage"`
```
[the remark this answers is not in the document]
dprince: not sure how these slipped in but I will fix them all. Thanks!

api/v1beta1/cinderapi_types.go
dprince: fixed all of these in the latest rebase. @Akrog thanks!

PR #338, opened 2024-02-15
Support LVM

Akrog:
With this PR we complete the LVM support in cinder.

We revert the workaround we had for LVM where we would use the OCP
internal network to be able to access from other pods in the same node
since now we can achieve something better easily during the deployment
by changing how we deploy the operators:

```
cd ~/install_yamls
NETWORK_STORAGE_MACVLAN=true make openstack
```

This PR also changes the LVM samples so that they can be used in a real
deployments with 3 nodes.

This is done by labeling one of the nodes as the destination for the LVM
backend.

Related `install_yamls` PR: https://github.com/openstack-k8s-operators/install_yamls/pull/731

### DIFF
--- a/config/samples/backends/bases/lvm/lvm.yaml
+++ b/config/samples/backends/bases/lvm/lvm.yaml
@@ -1,3 +1,9 @@
+# This machine config creates a systemd unit that creates a file based LVM VG
+# called cinder-volumes and mounts it on each reboot.
+#
+# Even though the systemd unit will be deployed in ALL OCP master nodes it will
+# only really run on the nodes that have the "openstack.org/cinder-lvm"
+# label.
 apiVersion: machineconfiguration.openshift.io/v1
 kind: MachineConfig
 metadata:
@@ -17,19 +23,23 @@ spec:
         Contents:  |
           [Unit]
           Description=Cinder LVM losetup
-          DefaultDependencies=no
           Conflicts=umount.target
-          Requires=lvm2-monitor.service systemd-udev-settle.service
-          Before=local-fs.target umount.target
-          After=var.mount lvm2-monitor.service systemd-udev-settle.service
+          Requires=lvm2-monitor.service systemd-udev-settle.service kubelet.service
+          After=var.mount lvm2-monitor.service systemd-udev-settle.service kubelet.service
 
           [Service]
           Environment=VG_GB_SIZE=10G
           Environment=LOOPBACK_FILE=/var/home/core/cinder-volumes
-          Type=oneshot
-          ExecStart=bash -c "if [[ ! -e ${LOOPBACK_FILE} ]]; then /bin/truncate -s ${VG_GB_SIZE} ${LOOPBACK_FILE} && /sbin/vgcreate cinder-volumes `/sbin/losetup --show -f ${LOOPBACK_FILE}`; else /sbin/losetup -f ${LOOPBACK_FILE}; fi"
-          ExecStop=bash -c "/sbin/losetup -d `/sbin/losetup -j ${LOOPBACK_FILE} -l -n -O NAME`"
+          Environment=LABEL=openstack.org/cinder-lvm
+
+          # Waits until cluster is available by failing, then Restart kicks in after 5 seconds
+          ExecStartPre=oc --kubeconfig=/var/lib/kubelet/kubeconfig get node
+          ExecStart=bash -c "if [ \"$(oc --kubeconfig=/var/lib/kubelet/kubeconfig get node -l ${LABEL} -o name)\" = \"node/$(hostname)\" ]; then if [[ ! -e ${LOOPBACK_FILE} ]]; then /bin/truncate -s ${VG_GB_SIZE} ${LOOPBACK_FILE} && /sbin/vgcreate cinder-volumes `/sbin/losetup --show -f ${LOOPBACK_FILE}`; else /sbin/losetup -f ${LOOPBACK_FILE}; fi; fi"
+          ExecStop=bash -c "if vgdisplay cinder-volumes; then /sbin/losetup -d `/sbin/losetup -j ${LOOPBACK_FILE} -l -n -O NAME`; fi"
           RemainAfterExit=yes
 
+          Restart=on-failure
+          RestartSec=5
+
           [Install]
-          WantedBy=local-fs-pre.target
+          WantedBy=multi-user.target
--- a/config/samples/backends/lvm/iscsi/backend.yaml
+++ b/config/samples/backends/lvm/iscsi/backend.yaml
@@ -3,7 +3,9 @@
 # The correct way to configure and run the daemons is using `MachineConfig` as shown in `iscsid.yaml` and `multipathd.yaml`.
 # To create the LVM on CRC the sample also uses a `MachineConfig` using a systemd unit with file `lvm.yaml`
 #
-# ATTENTION: After applying this OpenShift nodes will reboot, because of the `MachineConfig` changes and will take a while to recover.
+# ATTENTION:
+# - Node where we want LVM to run needs to be labeled with openstack.org/cinder-lvm
+# - After applying this OpenShift nodes will reboot, because of the `MachineConfig` changes and will take a while to recover.
 
 apiVersion: core.openstack.org/v1beta1
 kind: OpenStackControlPlane
@@ -14,6 +16,8 @@ spec:
     template:
       cinderVolumes:
         lvm-iscsi:
+          nodeSelector:
+            openstack.org/cinder-lvm: ""
           networkAttachments:
           - storage
           customServiceConfig: |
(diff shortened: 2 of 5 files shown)


controllers/cindervolume_controller.go
```diff
@@ -774,22 +783,38 @@ func processCustomServiceConfig(customServiceConfig string) (bool, string) {
 		} else if strings.HasPrefix(token, "[") && strings.HasSuffix(token, "]") {
 			// Note the section name before looking for a volume_backend_name
 			sectionName = strings.Trim(token, "[]")
+
+			if isBackendSection && isLVM {
```
ASBishop, line 787: I could follow this because I wrote the original code, but it took me a while to remember the parsing rules. I think what would help here is the block from L787..L792 is processing the previous (if any) backend section. It may not be intuitive that you are basically post-processing the previous section immediately after identifying a NEW section on L783.

controllers/cindervolume_controller.go
```diff
@@ -774,22 +783,38 @@ func processCustomServiceConfig(customServiceConfig string) (bool, string) {
...
+				usesLVM = true
+				if targetIP != "" {
+					lvmTargetIP = targetIP
+				}
+			}
+			// Reset the target ip for the section
+			targetIP = ""
+			isBackendSection = false
+			isLVM = true
```
ASBishop, line 796: This could use a comment. Initializing isLVM to True handles the fact that LVM is the default driver, so if no volume_driver is detected (at L803) then you must conclude it's an LVM backend. And if a volume_driver is detected at L803, the isLVM flag is properly updated at L805.

controllers/cindervolume_controller.go
```diff
@@ -684,20 +685,22 @@ func (r *CinderVolumeReconciler) generateServiceConfigs(
...
-		// c-bak to connect to LVM volumes.
 		networkAttachmentAddrs := cinder.GetNetworkAttachmentAddrs(
 			instance.Namespace, instance.Spec.NetworkAttachments, instance.Status.NetworkAttachments)
 
+		// Configure target IP addresses using all addresses in the network
+		// attachments. This relies on the fact that the LVM backend can only
+		// have one replica.
 		if len(networkAttachmentAddrs) > 0 {
-			templateParameters["TargetSecondaryIpAddresses"] = strings.Join(networkAttachmentAddrs, ",")
+			templateParameters["TargetIpAddress"] = networkAttachmentAddrs[0]
```
ASBishop, line 697: Something strange happens in my environment. I can see the "target_ip_address" properly set in the [backend_defaults] section in the 01-service-defaults.conf file. However, the logs clearly show the service is still using the OpenShift SDN address. This is weird. And if I manually patch my LVM CR to add the same target_ip_address to the backend (so it ends up in 03-service-custom.conf) then things seem to work. The debug logs at service startup still show the SDN address for the backend_defaults, and the proper storage network address for the LVM backend. I see these logs for the DEFAULT, backend_defaults and lvm sections:
```
target_ip_address              = 10.217.0.221
backend_defaults.target_ip_address = 10.217.0.221          <==== Wrong!
lvm.target_ip_address          = 172.18.0.33
```

ASBishop (changes requested):
In addition to the comments I left, I can't get glance to use cinder for its backend. Even after fixing the cinder-rtstool wrapper and configuring the proper target_ip_address, I see glance connect to the correct target address, but it triggers the dreaded "Image status transition from saving to saving is not allowed" error.

templates/cinder/config/cinder-volume-lvm-host-config.json
```diff
@@ -0,0 +1,47 @@
...
+    },
+    {
+      "source": "/usr/local/bin/container-scripts/run-on-host",
+      "dest": "/usr/sbin/nvme",
+      "owner": "root:root",
+      "perm": "0755"
+    },
+    {
+      "source": "/usr/local/bin/container-scripts/cinder-rtstool",
+      "dest": "/usr/sbin/cinder-rtstool",
```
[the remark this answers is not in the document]
Akrog: Since we have removed this approach and decided to only go with the `nncp` approach, this discussion is now irrelevant.

pkg/cindervolume/volumes.go
```diff
@@ -31,7 +32,14 @@ func GetVolumes(parentName string, name string, extraVol []cinderv1beta1.CinderE
 }
 
 // GetVolumeMounts - Cinder Volume VolumeMounts
-func GetVolumeMounts(name string, extraVol []cinderv1beta1.CinderExtraVolMounts) []corev1.VolumeMount {
+func GetVolumeMounts(name string, hostLVM bool, extraVol []cinderv1beta1.CinderExtraVolMounts) []corev1.VolumeMount {
+	var subPathCfg string
+	if hostLVM {
```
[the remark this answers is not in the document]
Akrog: Patch was removed comments no longer relevant.

controllers/cindervolume_controller.go
Akrog: Patch was removed comments no longer relevant.

config/samples/backends/bases/lvm/lvm.yaml
```diff
@@ -3,6 +3,7 @@ kind: MachineConfig
 metadata:
   labels:
     machineconfiguration.openshift.io/role: master
+    # kubernetes.io/hostname: "<hostname>"
```
ASBishop, line 6: If I understand the situation correctly, this might be a good place to add a comment, perhaps something like this:

```
For single node OCP, set the machineconfiguration.openshift.io/role to master. For 3-node OCP, change the role from master to worker, and specify the kubernetes.io/hostname  where you want the LVM backend data
to reside.
```
 
Alternatively, would it be easier to add a requirement to add a label (maybe "cinder-lvm") to an OCP node (regardless of single vs. 3-node OCP), and reference the label in the nodeSelector?

config/samples/backends/lvm/iscsi/backend.yaml
```diff
@@ -14,6 +18,9 @@ spec:
     template:
       cinderVolumes:
         lvm-iscsi:
+          nodeSelector:
```
ASBishop, line 21: Similar to my comment above, maybe this is a good place to add a YAML comment that explains what to do depending on whether you have a single or 3-node OCP.

config/samples/backends/lvm/nvme-tcp/backend.yaml
```diff
@@ -13,6 +17,9 @@ spec:
     template:
       cinderVolumes:
         lvm-nvme-tcp:
+          nodeSelector:
```
ASBishop, line 20: Same again.

controllers/cindervolume_controller.go
ASBishop: I no longer see this behavior in my latest tests. I don't know what happened when I observed the strange behavior, and if something was actually broken then it would be in cinder or oslo_config, and NOT the cinder-operator. I'm going to resolve this conversation so it doesn't confuse this PR.

config/samples/backends/lvm/iscsi/backend.yaml
```diff
@@ -3,7 +3,11 @@
 # The correct way to configure and run the daemons is using `MachineConfig` as shown in `iscsid.yaml` and `multipathd.yaml`.
 # To create the LVM on CRC the sample also uses a `MachineConfig` using a systemd unit with file `lvm.yaml`
 #
-# ATTENTION: After applying this OpenShift nodes will reboot, because of the `MachineConfig` changes and will take a while to recover.
+# ATTENTION:
+# - After applying this OpenShift nodes will reboot, because of the `MachineConfig` changes and will take a while to recover.
+# - If OCP has more than one node then:
+#   * Update `nodeSelector` here needs to be updated to use the hostname as the node selector.
```
Akrog, line 9: This reads wrong, probably better:
```
# * Update `nodeSelector` section in this file, it needs to use the chosen OCP node hostname.
```

config/samples/backends/lvm/iscsi/backend.yaml
```diff
@@ -3,7 +3,11 @@
...
+#   * Update the `99-master-cinder-lvm-losetup` `MachineConfig` to use the hostname as the label to select node.
```
Akrog, line 10: Also reads weird:
```
# * Update `99-master-cinder-lvm-losetup` to use the chosen OCP node hostname as one of the labels.
```

config/samples/backends/bases/lvm/lvm.yaml
ASBishop: Ah, thanks for clarifying. So you only use "workers" when there are >3 nodes.

Akrog:
> Looks great (I tested it, too!). I have one thought to consider before approval.
> 
> Currently the label is "openstack.org/cinder-backend=lvm" but an alternative might be "openstack.org/cinder-lvm-backend=true"
> 
> My thinking is "cinder-backend = lvm" might falsely imply there may be a need to label nodes when using other backends. I can imagine someone asking if they need a nodeSelector for "cinder-backend = XXX". Of course we know they'll need to label nodes for FC HBA, but the point is they don't need a label for every backend.

I have changed the label to "openstack.org/cinder-lvm"

ASBishop (approved):
/lgtm

PR #348, opened 2024-02-20
migrate from databaseUsername to databaseAccount and fully use MariaDBAccount

zzzeek:
This moves cinder to fully use MariaDBAccount based on the dev work being done for mariadb-operator:

https://github.com/openstack-k8s-operators/mariadb-operator/pull/184/files

Lead Jira: [OSPRH-4095](https://issues.redhat.com/browse/OSPRH-4095)

1. [x] controller calls EnsureMariaDBAccount up front to make sure MariaDBAccount is present
2. [x] error code from EnsureMariaDBAccount is handled, Conditions are amended when error is returned
3. [x] controller calls NewDatabaseForAccount instead of NewDatabase
4. [x]  GetAccountAndSecret is used to retrieve account /secret to populate template
5. [x]  GetDatabaseByName() , normally used for delete finalizers, replaced with GetDatabaseByNameAndAccount
6. [x]  CreateOrPatchAll() used to patch the Database, replacing CreateOrPatchDB / CreateOrPatchDBByName
7. [x]  controller calls DeleteUnusedMariaDBAccountFinalizers when launched pods are definitely running on a new MariaDBAccount, returns error code if present
8. [x]  PasswordSelectors that refer to database are removed
9. [x]  all databaseUser replaced with databaseAccount inside of all XYZ_types.go
10. [x] all databaseUser replaced with databaseAccount inside of all `kuttl/*.yaml`
11. [x] all default databaseUser names become databaseAccount, replacing underscores with dashes inside of all XYZ_types.go
12. [x] all default databaseUser names become databaseAccount, replacing underscores with dashes inside of all `kuttl/*.yaml`
13. [x] MariaDBAccountSuiteTests are used in controller ginkgo tests if it has them
14. [x] [184](https://github.com/openstack-k8s-operators/mariadb-operator/pull/184) is merged and replaces from go.mod are removed

### DIFF
--- a/controllers/cinder_controller.go
+++ b/controllers/cinder_controller.go
@@ -349,7 +349,7 @@ func (r *CinderReconciler) reconcileDelete(ctx context.Context, instance *cinder
 	Log.Info(fmt.Sprintf("Reconciling Service '%s' delete", instance.Name))
 
 	// remove db finalizer first
-	db, err := mariadbv1.GetDatabaseByName(ctx, helper, instance.Name)
+	db, err := mariadbv1.GetDatabaseByNameAndAccount(ctx, helper, instance.Name, instance.Spec.DatabaseAccount, instance.Namespace)
 	if err != nil && !k8s_errors.IsNotFound(err) {
 		return ctrl.Result{}, err
 	}
@@ -826,6 +826,11 @@ func (r *CinderReconciler) reconcileNormal(ctx context.Context, instance *cinder
 	instance.Status.Conditions.MarkTrue(condition.CronJobReadyCondition, condition.CronJobReadyMessage)
 	// create CronJob - end
 
+	err = mariadbv1.DeleteUnusedMariaDBAccountFinalizers(ctx, helper, instance.Name, instance.Spec.DatabaseAccount, instance.Namespace)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
 	Log.Info(fmt.Sprintf("Reconciled Service '%s' successfully", instance.Name))
 	return ctrl.Result{}, nil
 }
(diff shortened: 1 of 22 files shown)


controllers/cinder_controller.go
```diff
@@ -383,19 +383,17 @@ func (r *CinderReconciler) reconcileInit(
...
+		instance.Name,                  // CR name for MariaDBDatabase
+		instance.Spec.DatabaseAccount,  // CR name for MariaDBAccount
+		instance.Namespace,             // namespace
 	)
 	// create or patch the DB
-	ctrlResult, err := db.CreateOrPatchDB(
-		ctx,
-		helper,
-	)
+	// create or patch the DB
```
ASBishop, line 394: nit: duplicate of L393

ASBishop (commented):
+1 from me because CI shows it working

Discussions seem to be ongoing in https://github.com/openstack-k8s-operators/mariadb-operator/pull/184, so it would help to know when everything has settled and it's OK to start merging other operator PRs like this one.

controllers/cinder_controller.go
```diff
@@ -1242,24 +1251,43 @@ func (r *CinderReconciler) ensureDB(
...
+	)
+
+	db := mariadbv1.NewDatabaseForAccount(
+		instance.Spec.DatabaseInstance, // mariadb/galera service to target
+		instance.Name,                  // name used in CREATE DATABASE in mariadb
+		instance.Name,                  // CR name for MariaDBDatabase
+		instance.Spec.DatabaseAccount,  // CR name for MariaDBAccount
+		instance.Namespace,             // namespace
 	)
+	// create or patch the DB
```
zzzeek, line 1287: merge artifact is still there :(

controllers/cinder_controller.go
```diff
@@ -872,7 +877,7 @@ func (r *CinderReconciler) generateServiceConfigs(
 
 	labels := labels.GetLabels(instance, labels.GetGroupLabel(cinder.ServiceName), serviceLabels)
 
-	db, err := mariadbv1.GetDatabaseByName(ctx, h, cinder.DatabaseName)
+	db, err := mariadbv1.GetDatabaseByNameAndAccount(ctx, h, instance.Name, instance.Spec.DatabaseAccount, instance.Namespace)
```
ASBishop, line 880: Why is this necessary?
- Isn't the db supplied as an input (L870)? L880 simply supersedes the input param.
- If the caller isn't supplying the correct 'db' object then it be fixed in the caller's code.

ASBishop (changes requested):

controllers/cinder_controller.go
zzzeek: sure, the presence of `GetDatabaseByName` here should have been removed as part of #349 which rearranged the flow to pass `db` in.  So I can remove it here and make sure tests are not impacted.

PR #350, opened 2024-02-21
Remove notifications TODO

Akrog:
This patch updates our global default configuration that had a TODO for the cinder notifications to explicitly state that it is disabled by default and that it's very easy to enable it.

This is an excerpt of what a human operator would need to add to enable it:

```
apiVersion: core.openstack.org/v1beta1
kind: OpenStackControlPlane
spec:
  cinder:
    customServiceConfig: |
      [oslo_messaging_notifications]
      driver = messagingv2
      transport_url = {{ .TransportURL }}
```

Jira: #[OSPRH-1613](https://issues.redhat.com//browse/OSPRH-1613)

### DIFF
--- a/templates/cinder/config/00-global-defaults.conf
+++ b/templates/cinder/config/00-global-defaults.conf
@@ -45,10 +45,11 @@ lock_path = /var/locks/openstack/os-brick
 [oslo_concurrency]
 lock_path = /var/locks/openstack/cinder
 
-# TODO:
-#[oslo_messaging_notifications]
-#transport_url = rabbit://stackrabbit:secret@192.168.122.19:5672/
-#driver = messagingv2
+# Notifications are disabled by default, can be enabled by setting the driver
+# to messagingv2 at the top level cinder customServiceConfig
+[oslo_messaging_notifications]
+transport_url = {{ .TransportURL }}
+driver = noop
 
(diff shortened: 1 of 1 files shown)


templates/cinder/config/00-global-defaults.conf
```diff
@@ -45,9 +45,11 @@ lock_path = /var/locks/openstack/os-brick
 [oslo_concurrency]
 lock_path = /var/locks/openstack/cinder
 
-# TODO:
+# Notifications are disabled by default, they can be easily enabled by a human
```
ASBishop, line 48: FWIW, tripleo configures this with driver=noop. We could do the same here, and ask the operator to just override the driver. Configuring the transport_url here isn't a huge help to the operator because "{{ .TransportURL }}" makes it easy for them to configure it.
That said, if we intend to have the operator configure both the driver and transport_url in their customServiceConfig, then do we need to leave these comments in the global defaults file at all?

Akrog: Even if setting the `transport_url` is not a great help, I do think it's convenient, so I'll change this to a `noop`.
And writing this comment here was to make sure we had it documented in our repository that disabling it was on purpose and how to configure it in a convenient way.

PR #353, opened 2024-03-01
Restart services when their NodeSelector changes

ASBishop:
Add a hash of a service's Spec.NodeSelector to the inputs that determine when the corresponding pod should be recreated. This ensures pods are restarted whenever the NodeSelector is modified.


Akrog (approved):
/lgtm

fmount:
+1 from me as well, we need to trigger a reconciliation when this parameter change, otherwise is not useful.

fmount (approved):

PR #357, opened 2024-03-08
Revert "Restart services when their NodeSelector changes"

ASBishop:
This reverts commit df33f3ef495813f393a1cc9fc18555118463b9a1.

PR #353 was thought to be the solution to a problem observed while testing updates to how we manage cinder's LVM backend. This pertained to the use of a nodeSelector label that pins the c-vol pod to the k8s node where the LVM backend data is stored. Unfortunately, two subsequent issues were uncovered.

First, the nodeSelector already affects a service's statefulset, which in turn causes pods to be restarted automatically, There's no need to add the nodeSelector to the collection of hashes we use to restart pods when their configuration changes.

Second, the actual problem scenario we hoped to solve was getting a pod to be recreated when it got stuck in the Pending state due to a bad nodeSelector value. Currently we use an "Ordered" pod management policy, which means a Pending pod will need to be manually deleted. Hashing the nodeSelector has no impact on pods stuck in the Pending state.

### DIFF
--- a/controllers/cinderapi_controller.go
+++ b/controllers/cinderapi_controller.go
@@ -707,20 +707,6 @@ func (r *CinderAPIReconciler) reconcileNormal(ctx context.Context, instance *cin
 	instance.Status.Conditions.MarkTrue(condition.TLSInputReadyCondition, condition.InputReadyMessage)
 
 	//
-	// Hash the nodeSelector so the pod is recreated when it changes
-	//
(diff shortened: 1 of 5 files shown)


fmount (approved):
/lgtm

fmount:
after a conversation w/ @ASBishop I think we can approve this revert

PR #361, opened 2024-03-13
Fix LVM backend

Akrog:
This PR has 2 patches.

In the first one it fixes a couple of issues with the LVM backend:

- When the cinder-volume service is restarted for any reason
  (configuration change, update, etc) all connections to attached
  volumes will be broken and will never be recovered.

- At least with NVMe-oF (haven't checked with iSCSI), if a volume has
  been attached at any point and the cinder-volume service is restarted,
  then all following attachments won't work.

This patch fixes this issue by using the host network when using the LVM
backend.

Only the cinder-volume service is launched using the host's network
namespace so that the probes can still be run in an isolated network
namespaces; otherwise we would either need to disable the probes or
change the probes for LVM backends.

In the second patch the LVM samples are changed to use multipathing
using the tenant network as an additional one.


config/samples/backends/lvm/nvme-tcp/backend.yaml
```diff
@@ -28,5 +26,9 @@ spec:
             target_port=4420
             target_prefix=nvme-subsystem-1
             nvmeof_conn_info_version=2
-            use_multipath_for_image_xfer=false
+            use_multipath_for_image_xfer=true
```
ASBishop, line 29: L29 can be removed because 01-service-defaults.conf sets it to True.

config/samples/backends/lvm/iscsi/backend.yaml
```diff
@@ -28,3 +26,7 @@ spec:
             target_protocol=iscsi
             target_helper=lioadm
             volume_backend_name=lvm_iscsi
+            # This is the OCP host's static IP on the storage VLAN network
+            target_ip_address=172.18.0.5
```
ASBishop, line 30: Are these "magic" addresses when using CRC? What about 3-node OCP?

Akrog:
Added the WIP label because I want to double check this with the iSCSI backend, as I have only tested it with NVMe-TCP.

config/samples/backends/lvm/iscsi/backend.yaml
Akrog: Yes, those are for CRC when deploying with `install_yamls`, I think I mention it in the `README.md` file:
- https://github.com/openstack-k8s-operators/install_yamls/blob/17f749afc85155faba76f8fa878c213e7678f159/scripts/gen-nncp.sh#L244
- https://github.com/openstack-k8s-operators/install_yamls/blob/17f749afc85155faba76f8fa878c213e7678f159/scripts/gen-nncp.sh#L289

I don't know about `ci-framework` deployments.

PR #363, opened 2024-03-18
Add Cinder API scaled KUTTL tests

vhariria:


test/kuttl/tests/cinder_scale/02-scale-up-cinder-api.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kuttl.dev/v1beta1
+kind: TestStep
+commands:
+  - script: |
+      oc patch cinder -n $NAMESPACE  cinder --type='json' -p=\
```
ASBishop, line 5: I think you have too many references to "cinder" on this line. See the error message at test/kuttl/tests/cinder_scale

test/kuttl/tests/cinder_scale/02-scale-up-cinder-api.yaml
```diff
@@ -0,0 +1,6 @@
...
+       '[{"op": "replace", "path": "/spec/cinderAPI/replicas", "value":1}]'
```
ASBishop, line 6: Don't you want to set the value to 3?

test/kuttl/tests/cinder_scale/02-assert.yaml
```diff
@@ -0,0 +1,205 @@
...
+        name: config-data-custom
+      - emptyDir: {}
+        name: logs
+  updateStrategy:
+    rollingUpdate:
+      partition: 0
+    type: RollingUpdate
+status:
+  availableReplicas: 3
+  replicas: 3
```
ASBishop, line 170: I think this is the essence of the test. With this in mind, it would be good if you could trim the "assert" data down to the minimum, and not repeat everything that's already tested in the 01-assert.yaml (which is a symlink to ../../common/assert_sample_deployment.yaml.

The reason I'd like to see this 02-assert.yaml just test the replicas is it reduces the burden of maintaining all the assert files when something fundamental changes. We should be able to rely on 01-assert.yaml testing a lot of the deployment, and subsequent 02..0N-assert.yaml should just need to test the deltas.

ASBishop (changes requested):

test/kuttl/tests/cinder_scale/02-scale-up-cinder-api.yaml
vhariria: Ack, the test passed with 3 replicas in my local env. Updated the patch per local tests

test/kuttl/tests/cinder_scale/02-assert.yaml
vhariria: Thanks for the feedback, I will look into it.

vhariria: Ack, the trimmed file version worked in my setup earlier today. I will update the PR shortly. Thanks for the follow up.

ASBishop (approved):
/lgtm

PR #364, opened 2024-03-19
Fix Conditions

Akrog:
The way cinder-operator is handling conditions is not correct.

In this PR we make all necessary changes so that it behaves like
the minor update code expects and how a human operator would expect.

At the beginning of the reconcile we should always set all conditions to
Unknown to make sure that they are reset to the real status.

This is the right thing to do because we are doing the steps
sequentially, so any step we don't reach, we don't know its status,
hence the "Unknown" status condition.

Fix the global "Ready" condition for Cinder CRs so that it's only set to
"True" when we have gone through all the steps and not just based on the
current value.

The patch also set the global "Ready" condition to False instead of to
Unknown, as that's the case unless proven otherwise by reaching the end
of the reconciliation loop.

This is important for updating configuration of deployments, where we
may already have a running pod and the check that that pod is running
doesn't mean that the latest changes have been applied.

This PR also sets the status of some conditions that were not being
set and where we were just returning.

### DIFF
--- a/controllers/cinder_controller.go
+++ b/controllers/cinder_controller.go
@@ -151,11 +151,6 @@ func (r *CinderReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 
 	// Always patch the instance status when exiting this function so we can persist any changes.
 	defer func() {
-		// update the overall status condition if service is ready
-		if instance.IsReady() {
-			instance.Status.Conditions.MarkTrue(condition.ReadyCondition, condition.ReadyMessage)
-		}
-
 		err := helper.PatchInstance(ctx, instance)
 		if err != nil {
 			_err = err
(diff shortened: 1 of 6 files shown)


controllers/cinderapi_controller.go
```diff
@@ -856,7 +891,14 @@ func (r *CinderAPIReconciler) reconcileNormal(ctx context.Context, instance *cin
 			instance.Status.ReadyCount,
 		)
 		if err != nil {
-			return ctrl.Result{}, err
+			error := fmt.Errorf("verifying API NetworkAttachments (%s) %w", instance.Spec.NetworkAttachments, err)
```
ASBishop, line 894: Isn't this handled already at L911 below?

controllers/cinder_controller.go
```diff
@@ -627,6 +624,8 @@ func (r *CinderReconciler) reconcileNormal(ctx context.Context, instance *cinder
 		}
 	}
 
+	instance.Status.Conditions.MarkTrue(condition.NetworkAttachmentsReadyCondition, condition.NetworkAttachmentsReadyMessage)
```
ASBishop, line 633: I'm confused. Why is the top cinder_controller concerning itself with CinderAPI's network attachments (and only CinderAPI's attachments), when the cinderapi_controller has its own code?

controllers/cinderapi_controller.go
ASBishop: oh, right, my bad! The logic is so squirrely (too many conditionals) and potential exit points.

controllers/cinder_controller.go
ASBishop: The comment on L610 explains this controller looks at network attachments because dbSync needs the same ones as CinderAPI.

ASBishop (approved):
/lgtm

PR #368, opened 2024-03-21
Add Cinder Backup & Scheduler scaled KUTTL tests

vhariria:


ASBishop (approved):
/lgtm

Though I have a couple of thoughts to consider for future enhancements.
- It will be interesting to test scaling services down to 0 replicas and then scale up to >0
- If we anticipate adding more tests then the cleanup-cinder and errors YAML files could be renamed with "99-" prefix

PR #373, opened 2024-03-27
Remove inline RestoreLastTransitionTimes func

dprince:
This was added to lib-common


fmount (approved):
/lgtm

ASBishop:
Whoops, thanks for catching this!

PR #375, opened 2024-03-29
Mirror ReadyCondition with the actual current condition

ASBishop:
Initialize the ReadyCondition to Unknown, and in the controller's defer function mirror the latest current condition. The updates the ReadyCondition with information about whatever condition is not yet ready.


fmount (approved):
/lgtm

PR #378, opened 2024-04-19
Update some backend samples

ASBishop:
Update extraMounts for mounting ceph client files to use the 'secret' volume type instead of 'projected'. The projected volume type is meant for situations where there are multiple sources that need to all be merged, but with ceph there's just a single source.

Update the NFS and Netapp's ontap NFS configurations to switch from using a shares-config file to using the driver's nas_host and nas_share_patch config options. The shares-config file is not considered best practice for situations where multiple shares need to be configured. Instead, it's better to configure multiple backends, one per share, so that all volume requests go through the scheduler.

The NFS configurations also now include a few more recommended settings:

  nfs_snapshot_support=true
  nas_secure_file_operations=false
  nas_secure_file_permissions=false

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,10 +86,8 @@ cli.
           - CinderBackup
           volumes:
           - name: ceph
-            projected:
-              sources:
-              - secret:
-                  name: ceph-client-conf
+            secret:
+              secretName: ceph-client-conf
           mounts:
           - name: ceph
             mountPath: "/etc/ceph"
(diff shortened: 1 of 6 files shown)


fmount (approved):
/lgtm

README.md
```diff
@@ -151,10 +149,8 @@ spec:
         extraVolType: Ceph
         volumes:
         - name: ceph
-          projected:
-            sources:
-            - secret:
-                name: ceph-client-files
+          secret:
```
fmount, line 152: +1, no need for `projected` as it's the only secret we mount

PR #381, opened 2024-04-24
Fixing conditions

Akrog:
In this PR we fix the cinder operator to follow the latest recommendations on how to handle the status conditions of the CRDs.

At a high level the changes are:

- All CRDs have an `ObservedGeneration` field
- `ObservedGeneration` is updated with the `Generation` field value set by OpenShift as soon as the status conditions are initialized to `Unknown` in the beginning of the reconciliation.
- Top level `Cinder` CRD only considers sub CRDs to have valid data when their `Generation` and `ObservedGeneration` fields match.
- All controllers take into account their `ObservedGeneration` field to determine whether they are ready or not.

### DIFF
--- a/api/v1beta1/cinder_types.go
+++ b/api/v1beta1/cinder_types.go
@@ -177,7 +177,10 @@ type CinderStatus struct {
 	// ReadyCounts of Cinder Volume instances
 	CinderVolumesReadyCounts map[string]int32 `json:"cinderVolumesReadyCounts,omitempty"`
 
-	//ObservedGeneration - the most recent generation observed for this service. If the observed generation is less than the spec generation, then the controller has not processed the latest changes.
+	// ObservedGeneration - the most recent generation observed for this service.
+	// If the observed generation is different than the spec generation, then thei
+	// controller has not started processing the latest changes, and the status
+	// and its conditions are likely stale.
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 }
 
--- a/api/v1beta1/cinder_types.go
+++ b/api/v1beta1/cinder_types.go
@@ -226,7 +226,8 @@ func init() {
 
 // IsReady - returns true if all subresources Ready condition is true
 func (instance Cinder) IsReady() bool {
-	return instance.Status.Conditions.IsTrue(CinderAPIReadyCondition) &&
+	return instance.Generation == instance.Status.ObservedGeneration &&
+		instance.Status.Conditions.IsTrue(CinderAPIReadyCondition) &&
 		instance.Status.Conditions.IsTrue(CinderBackupReadyCondition) &&
 		instance.Status.Conditions.IsTrue(CinderSchedulerReadyCondition) &&
 		instance.Status.Conditions.IsTrue(CinderVolumeReadyCondition)
--- a/api/v1beta1/cinder_types.go
+++ b/api/v1beta1/cinder_types.go
@@ -178,7 +178,7 @@ type CinderStatus struct {
 	CinderVolumesReadyCounts map[string]int32 `json:"cinderVolumesReadyCounts,omitempty"`
 
 	// ObservedGeneration - the most recent generation observed for this service.
-	// If the observed generation is different than the spec generation, then thei
+	// If the observed generation is different than the spec generation, then the
 	// controller has not started processing the latest changes, and the status
 	// and its conditions are likely stale.
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
--- a/api/v1beta1/cinder_types.go
+++ b/api/v1beta1/cinder_types.go
@@ -160,13 +160,19 @@ type CinderStatus struct {
 	ServiceIDs map[string]string `json:"serviceIDs,omitempty"`
 
 	// ReadyCount of Cinder API instance
-	CinderAPIReadyCount int32 `json:"cinderAPIReadyCount,omitempty"`
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=0
+	CinderAPIReadyCount int32 `json:"cinderAPIReadyCount"`
 
 	// ReadyCount of Cinder Backup instance
-	CinderBackupReadyCount int32 `json:"cinderBackupReadyCount,omitempty"`
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=0
+	CinderBackupReadyCount int32 `json:"cinderBackupReadyCount"`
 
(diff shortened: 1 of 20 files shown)


controllers/cinder_controller.go
```diff
@@ -686,15 +688,18 @@ func (r *CinderReconciler) reconcileNormal(ctx context.Context, instance *cinder
 		Log.Info(fmt.Sprintf("Deployment %s successfully reconciled - operation: %s", instance.Name, string(op)))
 	}
 
-	// Mirror CinderAPI status' APIEndpoints and ReadyCount to this parent CR
-	instance.Status.APIEndpoints = cinderAPI.Status.APIEndpoints
-	instance.Status.ServiceIDs = cinderAPI.Status.ServiceIDs
-	instance.Status.CinderAPIReadyCount = cinderAPI.Status.ReadyCount
+	// Wait until the data in the StatefulSet is for the current generation
```
ASBishop, line 691: I like this comment (it's concise and informative). I think it can be repeated for other services as I note below.

controllers/cinder_controller.go
```diff
@@ -712,13 +717,15 @@ func (r *CinderReconciler) reconcileNormal(ctx context.Context, instance *cinder
 		Log.Info(fmt.Sprintf("Deployment %s successfully reconciled - operation: %s", instance.Name, string(op)))
 	}
 
-	// Mirror CinderScheduler status' ReadyCount to this parent CR
-	instance.Status.CinderSchedulerReadyCount = cinderScheduler.Status.ReadyCount
+	if cinderScheduler.Generation == cinderScheduler.Status.ObservedGeneration {
```
ASBishop, line 720: Repeat the comment at L691?

controllers/cinder_controller.go
```diff
@@ -741,24 +748,25 @@ func (r *CinderReconciler) reconcileNormal(ctx context.Context, instance *cinder
 			Log.Info(fmt.Sprintf("Deployment %s successfully reconciled - operation: %s", instance.Name, string(op)))
 		}
 
-		// Mirror CinderBackup status' ReadyCount to this parent CR
-		instance.Status.CinderBackupReadyCount = cinderBackup.Status.ReadyCount
+		// Only if the generation of the CinderBackup is the same as the observed generation, can we mirror the condition and ready count
```
ASBishop, line 751: This comment is worded very different from the one on L691.

controllers/cinder_controller.go
```diff
@@ -782,27 +791,32 @@ func (r *CinderReconciler) reconcileNormal(ctx context.Context, instance *cinder
...
-		}
-		instance.Status.CinderVolumesReadyCounts[name] = cinderVolume.Status.ReadyCount
-
-		// If this cinderVolume is not IsReady, mirror the condition to get the latest step it is in.
-		// Could also check the overall ReadyCondition of the cinderVolume.
-		if !cinderVolume.IsReady() {
-			c = cinderVolume.Status.Conditions.Mirror(cinderv1beta1.CinderVolumeReadyCondition)
-			// Get the condition with higher priority for volumeCondition.
-			volumeCondition = condition.GetHigherPrioCondition(c, volumeCondition).DeepCopy()
+		// Only if the generation of the CinderVolume is the same as the observed generation, can we mirror the condition and ready count
```
ASBishop, line 794: Another instance that might want to repeat L691.

api/v1beta1/cinder_types.go
```diff
@@ -160,18 +160,27 @@ type CinderStatus struct {
...
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=0
+	CinderSchedulerReadyCount int32 `json:"cinderSchedulerReadyCount"`
 
 	// ReadyCounts of Cinder Volume instances
 	CinderVolumesReadyCounts map[string]int32 `json:"cinderVolumesReadyCounts,omitempty"`
 
-	//ObservedGeneration - the most recent generation observed for this service. If the observed generation is less than the spec generation, then the controller has not processed the latest changes.
+	// ObservedGeneration - the most recent generation observed for this service.
+	// If the observed generation is different than the spec generation, then thei
```
ASBishop, line 181: nit: last word on this line has an extraneous 'i' ("thei")

api/v1beta1/cinderapi_types.go
```diff
@@ -136,5 +144,8 @@ func init() {
 
 // IsReady - returns true if service is ready to serve requests
 func (instance CinderAPI) IsReady() bool {
-	return instance.Status.ReadyCount == *instance.Spec.Replicas
+	return instance.Generation == instance.Status.ObservedGeneration &&
+		instance.Status.ReadyCount == *instance.Spec.Replicas &&
+		(instance.Status.Conditions.IsTrue(condition.DeploymentReadyCondition) ||
```
ASBishop, line 149: I can't tell whether L149 and L150 are truly necessary, or if they're already covered by L147 and L148. Are their situations where L147 and L148 are insufficient?

The same applies to the backup, sheduler and volume versions of the IsReady() function.

ASBishop (commented):
I have a few questions and noted a nit, but I don't have any technical concerns.

api/v1beta1/cinder_types.go
Akrog: Fixing it in next update.

controllers/cinder_controller.go
Akrog: I think I'll change it to reflect that it's not waiting, it's actually only updating things when the current generation matches, because I don't want to give the impression that this doesn't progress from this point forward if they don't match.

controllers/cinder_controller.go
Akrog: Done.

controllers/cinder_controller.go
Akrog: Changed.

controllers/cinder_controller.go
Akrog: Done.

api/v1beta1/cinderapi_types.go
Akrog: There are cases were the generation and the ready count matches but we should not be ready.

For example if you just update the configuration of a service. In that case as soon as you enter the reconciliation loop we make the generation match, and because the replicas haven't changed this method would incorrectly return `true`.

I believe we need those 2 extra conditions to make sure we always return the right value.

ASBishop: ack, that makes sense now.

ASBishop (approved):
/lgtm

PR #393, opened 2024-05-29
Wait for DB writes only on reads (causality checks)

Akrog:
On PR 392 we changed to waiting on writes for updates and inserts, but upon further analysis we have decided that just doing it for reads should be enough, as the order of the updates will be reconciled automatically by the DB engine.

In this patch we change the value of `mysql_wsrep_sync_wait` to 1 to only wait on reads.


zzzeek (approved):
let's just try it.  if we still have errors,  we can try adding bits though i dont think that would be the issue even then, we want to see that errors go away with the "1" setting

PR #395, opened 2024-06-05
Do not configure cinder's default_availability_zone

ASBishop:
Configure cinder's storage_availability_zone but leave the default_availability_zone unset so that the user can override the default "nova" value using either parameter.

For background, cinder has separate default_availability_zone and storage_availability_zone parameters, and the former supersedes the latter. If the default_availability_zone is unset, cinder will fallback to the storage_availability_zone. So, if the cinder operator configures the default_availability_zone then a cloud admin would be unable to override it with the
storage_availability_zone parameter. Historically, tripleo only configured the storage_availability_zone.


Akrog (approved):
/lgtm

PR #396, opened 2024-06-06
Add configurable API Timeouts

Akrog:
Multiple limitations exist with existing code:

- Human operator cannot configure Apache Timeout, so it's fixed at the defaults of 60 seconds.

  This can be problematic for cinder backends that under load can take longer on the export and mapping.

- Default HAProxy timeout is 30 seconds, which is different than the default in OSP17 (60 seconds)

- Human operators would need to sync values between 3 places: HAProxy, Apache, oslo.messaging sync RPC (call).

This patch adds an `apiTimeout` field to the `CinderSpecBase` to allow human operators to simultaneously configure the timeouts for HAProxy, Apache, and the `rpc_response_timeout`.

The `apiTimeout` defaults to 60 seconds, to mimic the behavior present in OSP17.

Having different timeouts for HAProxy, Apache, and `rpc_response_timeout` is possible setting the HAProxy timeout in the `apiOverride`, the Apache timeout with `apiTimeout`, and setting `rpc_response_timeout` in the top level Cinder `customServiceConfig`.

To be able to change the HAProxy value based on the `apiTimeout` with any update (and not just the first time) the code adds a custom annotation "api.cinder.openstack.org/timeout" with the value that was initially set, this way flags it as being set by the cinder-operator.

Jira: https://issues.redhat.com/browse/OSPRH-7393

### DIFF
--- a/api/bases/cinder.openstack.org_cinders.yaml
+++ b/api/bases/cinder.openstack.org_cinders.yaml
@@ -36,6 +36,10 @@ spec:
             type: object
           spec:
             properties:
+              apiTimeout:
+                default: 60
+                minimum: 10
+                type: integer
               cinderAPI:
                 properties:
                   containerImage:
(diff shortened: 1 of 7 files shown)


fmount:
looks good but I'll let @ASBishop approve

fmount (approved):
/lgtm

PR #397, opened 2024-06-06
Samples: iscsid login timeout

Akrog:
This patch changes the time to fail an iSCSI login request in our samples.

The default is 8 retries and 15 seconds for each one, so 120 seconds in total.

This patch changes the OpenShift cluster default to 3 retries and 5 seconds each (15 seconds in total), which is convenient for testing, as any healthy deployment and backend should be able to login to the backend in that amount of time, and if there is a broken path it will not take 2 minutes to give up.

Jira: https://issues.redhat.com/browse/OSPRH-7393
Jira: https://issues.redhat.com/browse/OSPRH-7415

### DIFF
--- a/config/samples/backends/bases/iscsid/iscsid.yaml
+++ b/config/samples/backends/bases/iscsid/iscsid.yaml
@@ -9,6 +9,22 @@ spec:
   config:
     ignition:
       version: 3.2.0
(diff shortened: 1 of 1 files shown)


fmount (approved):
/lgtm

PR #398, opened 2024-06-10
Set APITimeout defaults on the webhook

Akrog:
Kubebuilder doesn't always set the default of APITimeout, so with this patch we ensure that APITimeout field always has a valid value.


fmount (approved):
/lgtm

PR #400, opened 2024-06-12
Fix backend label when using non cinder name

Akrog:
When the Cinder CR is created using a name that is not "cinder" the `backend` label in the StatefulSet and the final cinder volume Pod will have the wrong value.

This is because the code that determines the backend name in the cinder volume controller doesn't have access to the actual name, so it tries to figure it out from the name of the CinderVolume CR, but since the name no longer starts with `cinder-volume`, then the value is wrong.

This patch changes the way we do this, and now the cinder-operator adds a label to the CinderVolume CR with the name of the backend, then the cinder volume controller uses it to pass it to the statefulset .

Jira: [OSPRH-7396](https://issues.redhat.com//browse/OSPRH-7396)

### DIFF
--- a/api/v1beta1/cindervolume_types.go
+++ b/api/v1beta1/cindervolume_types.go
@@ -22,6 +22,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	// Backend -
+	Backend = "backend"
+)
+
 // CinderVolumeTemplate defines the input parameters for the Cinder Volume service
 type CinderVolumeTemplateCore struct {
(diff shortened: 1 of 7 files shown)


Akrog:
recheck

gibizer (commented):
I suggest to use the new name pattern with a random UID in the functional test too:
 https://github.com/openstack-k8s-operators/cinder-operator/blob/3d0896ceb9a3c325261b4787041f57afdaf19ef4/test/functional/suite_test.go#L265

Akrog:
> I suggest to use the new name pattern with a random UID in the functional test too:

Done.

fmount (approved):
/lgtm

PR #401, opened 2024-06-18
Revert mysql wsrep sync wait

stuggi:
The openstack-k8s-operators/mariadb-operator#229 switched the mariadb-operator to deploy in Active/Passive mode per default, instead of multimaster.


Akrog (approved):
/lgtm

PR #413, opened 2024-07-08
Remove the cinder icon field from the CSV

Akrog:
There are legal reasons why the cinder project mascot cannot be used as an icon in the operator, so this patch removes it.

The icon is not really important, since it's the openstack-operator the one that should be important in the catalog.

Jira: [OSPRH-8413](https://issues.redhat.com//browse/OSPRH-8413)


fmount (approved):
/lgtm

PR #416, opened 2024-07-10
Update backend samples using custom images

Akrog:
Te way custom images are specified has changed, and now it no longer uses the `containerImage` file in the cinder sections but instead needs to be specified in the `OpenStackVersion`.

This patch updates the examples to match the latest code.


ASBishop (approved):
/lgtm

ASBishop (approved):
/lgtm

PR #417, opened 2024-07-12
Run cinder services as cinder user

eharney:
Run all cinder services as the cinder user instead of root.

This updates our kolla usage and apache configs to work in this configuration.

Jira: https://issues.redhat.com/browse/OSPRH-10083

### DIFF
--- a/pkg/cinderapi/statefuleset.go
+++ b/pkg/cinderapi/statefuleset.go
@@ -41,6 +41,7 @@ func StatefulSet(
 	annotations map[string]string,
 ) (*appsv1.StatefulSet, error) {
 	runAsUser := int64(0)
+	cinderUser := int64(cinderv1beta1.CinderUserID)
 
 	livenessProbe := &corev1.Probe{
 		// TODO might need tuning
@@ -156,7 +157,7 @@ func StatefulSet(
 							Args:  args,
 							Image: instance.Spec.ContainerImage,
 							SecurityContext: &corev1.SecurityContext{
-								RunAsUser: &runAsUser,
+								RunAsUser: &cinderUser,
 							},
 							Env:            env.MergeEnvs([]corev1.EnvVar{}, envVars),
 							VolumeMounts:   volumeMounts,
--- a/pkg/cinderapi/statefuleset.go
+++ b/pkg/cinderapi/statefuleset.go
@@ -30,7 +30,7 @@ import (
 
 const (
 	// ServiceCommand -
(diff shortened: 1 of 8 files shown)


ASBishop:
I tested a local, rebased version of this PR to determine whether changing the UID affects any operation or interoperability with other services that rely on file permissions. The good news is this is (nearly) a non issue.

The initial concern was the os-brick lock file directory that's shared between the cinder-volume, cinder-backup and glance-api control plane services. Consider the lock file named 'os-brick-connect_volume'. The file is created the first time any service acquires the lock, and the file will have the ownership and permissions of that service. After the lock is released, other services that need to acquire the lock must be able to open the file in write mode in order to take the lock, which in turn implies the other services need write access to the lock file. Prior to this PR the file could be created by a cinder service running as root, which should have caused problems for the glance-api service when running as the 'glance' user. Fortunately, tests show this doesn't happen, apparently as a positive side effect of k8s's UID and namespace remapping. Glance is able to acquire the lock even when the lock file's permissions suggest it doesn't have write permission. Further testing with this PR, with cinder services running as the 'cinder' user, there are no problems with any service acquiring the lock, regardless of who created the file (i.e. which service first acquired the lock).

The next concern pertains to NFS volumes, and whether volumes created when cinder was running as root would encounter a problem after switching to run as 'cinder'. This is a non-issue, because cinder's "remotefs" driver has provisions for managing the NFS volume's backing file's permissions. Switching from root to 'cinder' is seamless.

A final concern pertains to using an extraMount in order to perform image conversion operations on an NFS share provided by the extraMount. Tests show an issue does arise if the NFS share (on the server) doesn't allow non-root users to write to the share. This can be handled by documenting the issue. The NFS share's directory permission may need to be updated to grant write permission to the 'cinder' user (UID / GID 42407).

rajathere:
I did a fair bit of testing myself and tldr; everything seems to work fine.

Basic Test: After applying changes, pods should run as cinder user

Some advance tests:
1. Can cinder user create/acquire locks: SUCCESS
2. If glance creates a lock, can cinder acquire it and vice versa: SUCCESS
3. If glance acquires the lock, does cinder wait for lock to be released before proceeding with the operation: SUCCESS

So everything seems to work fine with this change.
One question i still have is how the different can access/create/delete locks in the HostPath directory on the host node even without the correct permissions but that  should not block this PR.
LGTM.

rajathere (approved):

PR #427, opened 2024-08-01
Correct formatting of ceph doc

rajathere:
The yaml block was not formatted within ```example``` causing it to be formatted as a paragraph in a single line.


Akrog (approved):
/lgtm

PR #446, opened 2024-09-25
NVMe sample: Fix nvme ids

Akrog:
OpenShift image has the NVMe `hostid` and `hostnqn` hardcoded, so it ends up being the same in all the control plane nodes, which is problematic.

This `MachineConfig` fixes this issue by recreating both files when the `hostid` doesn't match the system-uuid of the machine it is running on.

Bugs:
- https://issues.redhat.com/browse/OCPBUGS-34629
- https://github.com/openshift/os/issues/1519
- https://github.com/openshift/os/pull/1520

### DIFF
--- a/config/samples/backends/bases/nvmeof/fix-ocp-nvme.yaml
+++ b/config/samples/backends/bases/nvmeof/fix-ocp-nvme.yaml
@@ -1,0 +1,38 @@
+# The OpenShift image has the NVMe hostid and hostnqn hardcoded, so it ends up
+# being the same in all the control plane nodes, which is problematic.
+# This MachineConfig fixes this issue by recreating both files when the hostid
+# doesn't match the system-uuid of the machine it is running on.
+# Bugs:
+#   https://issues.redhat.com/browse/OCPBUGS-34629
+#   https://github.com/openshift/os/issues/1519
+#   https://github.com/openshift/os/pull/1520
+#
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    component: fix-nvme-ids
+    machineconfiguration.openshift.io/role: master
+    service: cinder
+  name: 99-master-cinder-fix-nvme-ids
+spec:
+  config:
+    Systemd:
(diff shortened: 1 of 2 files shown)


config/samples/backends/bases/nvmeof/fix-ocp-nvme.yaml
```diff
@@ -0,0 +1,38 @@
...
+#   https://issues.redhat.com/browse/OCPBUGS-34629
+#   https://github.com/openshift/os/issues/1519
+#   https://github.com/openshift/os/pull/1520
+#
+apiVersion: machineconfiguration.openshift.io/v1
+kind: MachineConfig
+metadata:
+  labels:
+    component: fix-nvme-ids
+    machineconfiguration.openshift.io/role: master
```
ASBishop, line 15: Sorry, I just spotted this now, but should the role be worker? I think our other machineConfig specify master nodes, and that works in our small environments where nodes are both.

Akrog: For our samples it should be `master`, like in all the other `MachineConfig`s we have in our samples.

It's only `worker` when we are deploying in a 3 master and N workers deployment and OpenStack control plane will be running on the workers.

This is described in the generic documents PR: https://github.com/openstack-k8s-operators/cinder-operator/blob/ff49b642d16e7537be26e38b9a5115d0b942b62b/docs/user-guide/cinder.md?plain=1#L156

```
> **⚠ Attention:** `MachineConfig` examples in the following sections are using
label `machineconfiguration.openshift.io/role: worker` instructing OpenShift to
apply those changes on `worker` nodes (`workers` is an automatically created
`MachineConfigPool`)  This assumes we have an OpenShift cluster with 3 master
nodes and 3 `worker` nodes.  If we are deploying in a 3 node OpenShift cluster
where all nodes are `master` and `worker` we need to change it to use `master`
instead (`master` is another automatically created `MachineConfigPool`).
```

ASBishop (approved):
/lgtm

PR #463, opened 2024-11-05
Fix extraMounts propagation when uniquePodNames is used

fmount:
When `uniquePodNames` is set to `true`, every Cinder `Pod` (but in general each component and service) is prefixed by a pseudo random string. This affects the per-instance propagation, because the old way based on `strings.TrimPrefix` is not valid anymore. This patch switches the implementation to a more solid approach: the instance name, stored as `Label` in `CinderVolume`, is retrieved using the associated `BackendName()` function.
By doing this (we do something similar in Glance already) we can easily retrieve the instance name and fix the `extraMounts` propagation.


abays (approved):
/lgtm

PR #466, opened 2024-11-15
Ensure nodeSelector logic is consistent for all operators

olliewalsh:
nodeSelectors are not currently applied consistently across all operators.

Some do not support nodeSelectors at all - will be implemented in this series of pull requests.
Some do not apply them to every pod (jobs/cronjobs esp) - will be resolved in this series of pull requests.
Some do not apply a node selector update correctly, only when not initially set.
Some support nodeSelectors but do not inherit the default nodeSelector from the OpenstackControlPlane CR.

Jira: [OSPRH-10734](https://issues.redhat.com//browse/OSPRH-10734)

### DIFF
--- a/test/functional/base_test.go
+++ b/test/functional/base_test.go
@@ -19,6 +19,7 @@ import (
 	"golang.org/x/exp/maps"
 
 	. "github.com/onsi/gomega" //revive:disable:dot-imports
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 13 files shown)


test/functional/base_test.go
```diff
@@ -230,6 +231,14 @@ func GetCinderVolume(name types.NamespacedName) *cinderv1.CinderVolume {
 	return instance
 }
 
+func GetCronJob(name types.NamespacedName) *batchv1.CronJob {
```
stuggi, line 234: in a follow up should add this to https://github.com/openstack-k8s-operators/lib-common/tree/main/modules/common/test/functional

olliewalsh: yes, also for keystone where this came from

stuggi (approved):
/lgtm

PR #476, opened 2024-12-10
Add extraMounts functional tests

fmount:
This change adds a functional test for `Cinder` `extraMounts`. It ensures we're able to validate the abstraction of `corev1.VolumeSource` struct introduced in `lib-common/storage` module.

Jira: https://issues.redhat.com/browse/OSPRH-11210

### DIFF
--- a/test/functional/base_test.go
+++ b/test/functional/base_test.go
@@ -296,3 +296,37 @@ func CinderVolumeNotExists(name types.NamespacedName) {
 		g.Expect(k8s_errors.IsNotFound(err)).To(BeTrue())
 	}, timeout, interval).Should(Succeed())
 }
+
+// GetExtraMounts - Utility function that simulates extraMounts pointing
+// to a Ceph secret
+func GetExtraMounts() []map[string]interface{} {
+	return []map[string]interface{}{
+		{
+			"name":   cinderTest.Instance.Name,
+			"region": "az0",
+			"extraVol": []map[string]interface{}{
+				{
+					"extraVolType": CinderCephExtraMountsSecretName,
+					"propagation": []string{
+						"CinderVolume",
+					},
+					"volumes": []map[string]interface{}{
+						{
+							"name": CinderCephExtraMountsSecretName,
+							"secret": map[string]interface{}{
+								"secretName": CinderCephExtraMountsSecretName,
+							},
+						},
+					},
+					"mounts": []map[string]interface{}{
+						{
+							"name":      CinderCephExtraMountsSecretName,
+							"mountPath": CinderCephExtraMountsPath,
+							"readOnly":  true,
+						},
+					},
+				},
+			},
+		},
+	}
+}
--- a/test/functional/cinder_controller_test.go
+++ b/test/functional/cinder_controller_test.go
@@ -863,6 +863,68 @@ var _ = Describe("Cinder controller", func() {
 
 	})
 
+	When("Cinder CR instance is built with ExtraMounts", func() {
+		BeforeEach(func() {
+			rawSpec := map[string]interface{}{
+				"secret":              SecretName,
+				"databaseInstance":    "openstack",
+				"rabbitMqClusterName": "rabbitmq",
+				"extraMounts":         GetExtraMounts(),
+				"cinderAPI": map[string]interface{}{
+					"containerImage": cinderv1.CinderAPIContainerImage,
+				},
+				"cinderScheduler": map[string]interface{}{
+					"containerImage": cinderv1.CinderSchedulerContainerImage,
+				},
+				"cinderVolumes": map[string]interface{}{
+					"volume1": map[string]interface{}{
+						"containerImage": cinderv1.CinderVolumeContainerImage,
+					},
+				},
+			}
+
+			DeferCleanup(th.DeleteInstance, CreateCinder(cinderTest.Instance, rawSpec))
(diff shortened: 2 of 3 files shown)


test/functional/cinder_controller_test.go
```diff
@@ -863,6 +863,75 @@ var _ = Describe("Cinder controller", func() {
...
+		It("Check the extraMounts of the resulting StatefulSets", func() {
+			th.SimulateStatefulSetReplicaReady(cinderTest.CinderAPI)
+			th.SimulateStatefulSetReplicaReady(cinderTest.CinderScheduler)
+			// Retrieve the generated resources
+			volume := cinderTest.CinderVolumes[0]
+			th.SimulateStatefulSetReplicaReady(volume)
+			ss := th.GetStatefulSet(volume)
+			// Check the resulting deployment fields
+			Expect(int(*ss.Spec.Replicas)).To(Equal(1))
+			Expect(ss.Spec.Template.Spec.Volumes).To(HaveLen(15))
```
stuggi, line 918: could also use https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/test/helpers/volumes.go#L27 . to check that the volume with the expected name exist.

test/functional/cinder_controller_test.go
```diff
@@ -863,6 +863,75 @@ var _ = Describe("Cinder controller", func() {
...
+			Expect(ss.Spec.Template.Spec.Containers).To(HaveLen(2))
+			// Get the manila-share container
+			container := ss.Spec.Template.Spec.Containers[1]
+			// Fail if manila-share doesn't have the right number of
+			// VolumeMounts entries
+			Expect(container.VolumeMounts).To(HaveLen(17))
```
stuggi, line 924: we have https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/test/helpers/volumes.go#L36 to check that an expected VolumeMount exist, but it does not (yet) provide functionality to validate the `MountPath`. should we enhance it?

test/functional/base_test.go
```diff
@@ -296,3 +296,37 @@ func CinderVolumeNotExists(name types.NamespacedName) {
 		g.Expect(k8s_errors.IsNotFound(err)).To(BeTrue())
 	}, timeout, interval).Should(Succeed())
 }
+
+// GetExtraMounts - Utility function that simulates extraMounts pointing
+// to a Ceph secret
+func GetExtraMounts() []map[string]interface{} {
```
stuggi, line 302: wondering if it would make sense and help others to add a test helper to the storage module in https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/storage or is it to specific for the component?

stuggi (commented):
looks good to me, just some general comments which are not something request to change. just some feedback or to get your opinion on.

fmount:
> looks good to me, just some general comments which are not something request to change. just some feedback or to get your opinion on.

Thank you very much for the review, it's very useful and I wasn't aware about existing assertions provided for basic components in lib-common.
I'm going to push an update to make sure we reuse functions like `AssertVolumes`/`AssertVolumeMounts` already provided by lib-common. This would simplify a bit the core logic here!

test/functional/cinder_controller_test.go
fmount: Yes, I totally agree to this. Thank you, I wasn't aware about lib-common providing such assertion, and I'm going to use it. In general I don't like much  checking the `Volumes` (or `VolumeMount`) array length by hardcoding a value, and the assertion solves this problem providing a more solid check!

test/functional/base_test.go
fmount: This function is specific to the component,it  only craft a piece of the CR that is applied beforehand, and later checked within a dedicated envTest.
I never thought about providing a `lib-common/modules/storage` helper, but this would definitely be an interesting idea that would help to:
1. simplify the operators test logic when need to validate an `extraMount`
2. provide a wrapper to perform the assertion and hide the detailed steps required to get the Volume/VolumeMount and call the existing assert

Perhaps I can do that in a follow up so I can try to add such helper in `lib-common` first.

test/functional/cinder_controller_test.go
fmount: This seems a good idea as well. I was planning, as part of the other comment, to add a validation  as part of the `extraMount.propagation`, to make sure the `VolumeMount` exists in the resulting statefulset, but perhaps when you mentioned `validate the MountPath` you had something different in mind (e.g. reserved paths or other relevant `RFCs`)

test/functional/base_test.go
stuggi: sure, doing this later in a follow up whenever time permits is fine

test/functional/cinder_controller_test.go
stuggi: in your previous version you were validating the volumemount path like

~~~
Expect(vm.MountPath).To(	ContainSubstring(CinderCephExtraMountsPath))
~~~

we could enhance the test helpers to be able to do that. thats not something it checks.

fmount: oh I see sorry. Even for `MountPath` we currently check the name and it would be good to be able to validate the path that way. I will add it to the existing assertion in `lib-common`, so we get it here with the next bump.

fmount:
@stuggi I have the `mountPath` check in my backlog. I'm not planning to change the signature of the original function as it would break existing tests, but I'm going to build a v2 function that includes the first one, and executes the `mountPath` check. I'll follow up with a lib-common change for it.
Can we land this patch in the meantime? (so I can close this first round of `extraMounts` test).

stuggi (approved):
sure, thought I already did. but missed to approve

PR #482, opened 2025-02-19
Consume Topology CR by reference

fultonj:
Allow Pod placement and scheduling to be made through the Topology CR integration introduced in [1].

- a new API parameter (TopologyRef) defined for each Component that allows to reference an existing Topology CRs in the same namespace

- the operator logic that retrieves and processes the referenced Topology CR through the functions provided by lib-common [2]

- StatefulSet configuration incorporates the processed Topology

- a set of envTest to test the lifecycle (add/update/override/remove) of the resulting StatefulSets when a Topology is referenced

webhooks are in place to prevent referencing a Topology from a different namespace (which is not supported).

This patch is based on [3] by fpantano@redhat.com

[1] https://github.com/openstack-k8s-operators/infra-operator/pull/325
[2] https://github.com/openstack-k8s-operators/lib-common/pull/594
[3] https://github.com/openstack-k8s-operators/manila-operator/pull/385

Jira: https://issues.redhat.com/browse/OSPRH-13274


fultonj:
obsoleted by https://github.com/openstack-k8s-operators/cinder-operator/pull/483

PR #500, opened 2025-04-14
Optmize handling topology logic

fmount:
This patch removes the need of using `GetLastAppliedTopologyRef` to build a `topologyRef` used by `EnsureServiceTopology` provided by infra-operator. In particular, we can now blindly pass `instance.Status.LastAppliedTopologyRef`, and checking that is not nil is deferred to the infra-operator logic [1]. This change also allows to deduplicate most of the common code present in all operators.
As part of the same change, envTests are improved and both `CreateTopology` and `GetTopology` function calls now rely on infra-operator's implementation. 

Depends-On: https://github.com/openstack-k8s-operators/infra-operator/pull/367
Jira: https://issues.redhat.com/browse/OSPRH-15412

[1] https://github.com/openstack-k8s-operators/infra-operator/pull/367


api/go.mod
```diff
@@ -72,3 +72,5 @@ replace github.com/openshift/api => github.com/openshift/api v0.0.0-202408300231
 
 // custom RabbitmqClusterSpecCore for OpenStackControlplane (v2.6.0_patches_tag)
 replace github.com/rabbitmq/cluster-operator/v2 => github.com/openstack-k8s-operators/rabbitmq-cluster-operator/v2 v2.6.1-0.20241017142550-a3524acedd49 //allow-merging
+
+replace github.com/openstack-k8s-operators/infra-operator/apis => github.com/fmount/infra-operator/apis v0.0.0-20250403071512-a2f98fcd7e19
```
fultonj, line 76: This will need to be removed before merging. I assume you knew, but just being explicit instead of implicit.

fultonj (commented):
Other than the reference to your personal repo, this looks good to me.

api/go.mod
fmount: Done

stuggi (approved):
/lgtm

PR #505, opened 2025-04-28
Enable MTLS memcached auth

lmiccini:
This commit allows operators to use mtls as an authentication method against Memcached.
Cinder controllers will detect the presence of a purposely-created mtls certificate (authCertSecret in the memcached spec) and use this to configure the [keystone_authtoken] section accordingly. Additional volumes/volumemounts will be added to each pod.

Note that this commit switches from MemcachedServersWithInet to MemcachedServers since keystone-middleware now uses pymemcache when tls=true and there is no need to use "[]" to enclose the list of memcached servers even for ipv6.


api/v1beta1/cinderapi_types.go
```diff
@@ -84,6 +84,11 @@ type CinderAPISpec struct {
 	// +kubebuilder:validation:Required
 	// ServiceAccount - service account name used internally to provide Cinder services the default SA name
 	ServiceAccount string `json:"serviceAccount"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=memcached
+	// Memcached instance name.
+	MemcachedInstance string `json:"memcachedInstance"`
```
fmount, line 91: I think this should be a `*string` to make sure we gracefully handle minor updates.
When is a *string, if this parameter doesn't exist, you won't trigger any Pod restart/rollout (you should just need to handle the code you already have with a `instance.Spec.MemcachedInstance != nil`.
By keeping the code as it is we risk to trigger a cinder restart when the operator is updated.

api/v1beta1/cinderbackup_types.go
```diff
@@ -72,6 +72,11 @@ type CinderBackupSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// TLS - Parameters related to the TLS
 	TLS tls.Ca `json:"tls,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=memcached
+	// Memcached instance name.
+	MemcachedInstance string `json:"memcachedInstance"`
```
fmount, line 79: same concern here, a new operator update will update the Spec with this parameter and trigger an unwanted restart for cinder backup. We should turn it into a *string to make sure we can omit it everywhere and be fully backward compatible,

api/v1beta1/cinderscheduler_types.go
```diff
@@ -72,6 +72,11 @@ type CinderSchedulerSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// TLS - Parameters related to the TLS
 	TLS tls.Ca `json:"tls,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=memcached
+	// Memcached instance name.
+	MemcachedInstance string `json:"memcachedInstance"`
```
fmount, line 79: same comment applies here

api/v1beta1/cindervolume_types.go
```diff
@@ -78,6 +78,11 @@ type CinderVolumeSpec struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec
 	// TLS - Parameters related to the TLS
 	TLS tls.Ca `json:"tls,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=memcached
+	// Memcached instance name.
+	MemcachedInstance string `json:"memcachedInstance"`
```
fmount, line 85: same comment applies here

controllers/cinder_controller.go
```diff
@@ -1021,6 +1029,7 @@ func (r *CinderReconciler) apiDeploymentCreateOrUpdate(ctx context.Context, inst
 		DatabaseHostname:   instance.Status.DatabaseHostname,
 		TransportURLSecret: instance.Status.TransportURLSecret,
 		ServiceAccount:     instance.RbacResourceName(),
+		MemcachedInstance:  instance.Spec.MemcachedInstance,
```
fmount, line 1032: in case we turn the {api,scheduler,volume,backup}.Spec.MemcachedInstance into a pointer (*string), we should change this logic to only assign it when `mTLS` is enabled.
In other words we should remove L1032, and have something like:

```golang
if memcached.GetMemcachedMTLSSecret() != "" {
  cinderAPISpec.MemcachedInstance = instance.Spec.MemcachedInstance
}

File: controllers/cinder_controller.go [hunk 6]
```diff
@@ -1063,6 +1072,7 @@ func (r *CinderReconciler) schedulerDeploymentCreateOrUpdate(ctx context.Context
 		TransportURLSecret:      instance.Status.TransportURLSecret,
 		ServiceAccount:          instance.RbacResourceName(),
 		TLS:                     instance.Spec.CinderAPI.TLS.Ca,
+		MemcachedInstance:       instance.Spec.MemcachedInstance,
```
fmount, line 1075: same comment I added to `cinderAPISpec`

controllers/cinder_controller.go
```diff
@@ -1105,6 +1115,7 @@ func (r *CinderReconciler) backupDeploymentCreateOrUpdate(ctx context.Context, i
 		TransportURLSecret:   instance.Status.TransportURLSecret,
 		ServiceAccount:       instance.RbacResourceName(),
 		TLS:                  instance.Spec.CinderAPI.TLS.Ca,
+		MemcachedInstance:    instance.Spec.MemcachedInstance,
```
fmount, line 1118: same comment I added to `cinderAPISpec`

controllers/cinder_controller.go
```diff
@@ -1168,6 +1179,7 @@ func (r *CinderReconciler) volumeDeploymentCreateOrUpdate(ctx context.Context, i
...
```
fmount, line 1182: same comment I added to `cinderAPISpec`

pkg/cinderbackup/statefulset.go
```diff
@@ -96,6 +98,12 @@ func StatefulSet(
 		volumeMounts = append(volumeMounts, instance.Spec.TLS.CreateVolumeMounts(nil)...)
 	}
 
+	// add MTLS cert if defined
+	if memcached.Status.MTLSCert != "" {
```
fmount, line 102: as per the comments I had, if `MemcachedInstance` is a *string in the sub resources (for minor updates backward compatibility), this condition should also depends on memcached (`if memcached.Status.MTLSCert != "" && instance.Spec.MemcachedInstance != nil`)

pkg/cinderscheduler/statefulset.go
```diff
@@ -86,6 +88,12 @@ func StatefulSet(
 		volumeMounts = append(volumeMounts, instance.Spec.TLS.CreateVolumeMounts(nil)...)
 	}
 
+	// add MTLS cert if defined
+	if memcached.Status.MTLSCert != "" {
```
fmount, line 92: as per the comments I had, if `MemcachedInstance` is a `*string` in the sub resources (for minor updates backward compatibility), this condition should also depends on memcached (`if memcached.Status.MTLSCert != "" && instance.Spec.MemcachedInstance != nil`)

pkg/cindervolume/statefulset.go
```diff
@@ -103,6 +105,12 @@ func StatefulSet(
 		volumeMounts = append(volumeMounts, instance.Spec.TLS.CreateVolumeMounts(nil)...)
 	}
 
+	// add MTLS cert if defined
+	if memcached.Status.MTLSCert != "" {
```
fmount, line 109: as per the comments I had, if `MemcachedInstance` is a `*string` in the sub resources (for minor updates backward compatibility), this condition should also depends on memcached (`if memcached.Status.MTLSCert != "" && instance.Spec.MemcachedInstance != nil`)

pkg/cinderapi/statefuleset.go
```diff
@@ -86,6 +88,12 @@ func StatefulSet(
 		volumeMounts = append(volumeMounts, instance.Spec.TLS.CreateVolumeMounts(nil)...)
 	}
 
+	// add MTLS cert if defined
+	if memcached.Status.MTLSCert != "" {
```
fmount, line 92: as per the comments I had, if `MemcachedInstance` is a `*string` in the sub resources (for minor updates backward compatibility), this condition should also depends on memcached (`if memcached.Status.MTLSCert != "" && instance.Spec.MemcachedInstance != nil`)

fmount (commented):
Hi @lmiccini,
thanks for starting this!
I had a first pass and a few comments (they're all related), based on a potential issue wrt minor updates, where we could end up triggering an unwanted rollout/restart in case the new cinder-operator version performs a reconciliation with the new subCRD.
I think the implementation is ok, but we might want to turn the new parameter into a *string to avoid this unwanted restart of all cinder resources.
We did this for topologies and we included a `*TopologyRef`, I think the same pattern applies here.
/cc @stuggi for any additional thought on this.

controllers/cinderbackup_controller.go
```diff
@@ -568,8 +570,13 @@ func (r *CinderBackupReconciler) reconcileNormal(ctx context.Context, instance *
 		return ctrl.Result{}, fmt.Errorf("waiting for Topology requirements: %w", err)
 	}
 
+	memcached, err := memcachedv1.GetMemcachedByName(ctx, helper, *instance.Spec.MemcachedInstance, instance.Namespace)
```
fmount, line 573: I'm wondering if this breaks the existing (`nonTLS`) use case.
Imagine we have a previous deployment (e.g. FR1) where there's no knowledge about `TLS`.
We update operators first, and we land to FR3 operators, where this code is present. Because that environment has no `TLS`,  `instance.Spec.MemcachedInstance` is not propagated to the underlying sub resources, and the pointer to that parameter is nil.
Now imagine you reach this point (the same applies to the other sub-resources (I'm not going to duplicate my comment), where `instance.Spec.MemcachedInstance` is `nil`: you can either segfault the operator because you're dereferencing a nil pointer, or (lucky case) an `err` is returned and you return it.
Does this makes sense? Do we need to check `memcached` when `instance.Spec.MemcachedInstance != nil`?

fmount: Technically we do not fail in the jobs because the parameter is added to the sub-resources, but it seems a side effect of `kubebuilder` enforcing it in a greenfield deployment[1], while the config is not updated. This is the reason we do not see any panic.
I'm wondering if it is acceptable that we document Pod restarts during minor updates or we should work more on it and not have this parameter at all injected in the `Spec`?

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_cinder-operator/505/pull-ci-openstack-k8s-operators-cinder-operator-main-cinder-operator-build-deploy-tempest/1918535446334279680/artifacts/cinder-operator-build-deploy-tempest/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-970d6d703a1bc138ac47064ac1411ad5b432d97f896b5fa1e6611d0a3cbfefb5/namespaces/openstack/crs/cinderapis.cinder.openstack.org/cinder-api.yaml

[2] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_cinder-operator/505/pull-ci-openstack-k8s-operators-cinder-operator-main-cinder-operator-build-deploy-tempest/1918535446334279680/artifacts/cinder-operator-build-deploy-tempest/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-970d6d703a1bc138ac47064ac1411ad5b432d97f896b5fa1e6611d0a3cbfefb5/namespaces/openstack/secrets/cinder/cinder-api-config-data.yaml-00-global-defaults.conf

lmiccini: that's one of the reasons why my initial approach was to "look up" and not pass down the memcached instance to all the resources, I'll see if I can figure out something to deal with the case where memcached = nil.

lmiccini: can't we just rely on the nil returned by https://github.com/openstack-k8s-operators/infra-operator/blob/2ce4eea3a06dcd15429e99acd261ec3f52630a89/apis/memcached/v1beta1/memcached_funcs.go#L136C3-L136C18 in case there is no instance.Spec.MemcachedInstance ?

fmount: That would probably be the case, and in that case you should be good!

PR #513, opened 2025-06-04
Watch KeystoneAPI status updates to reconcile

stuggi:
Adds watches of the Nova controller on the KeystoneAPI status to reconcile if e.g. the endpoint list changes. This triggers setting the new KeystoneAuthURL on the sub components to update their configuration.

Depends-On: https://github.com/openstack-k8s-operators/keystone-operator/pull/591

Jira: [OSPRH-16994](https://issues.redhat.com//browse/OSPRH-16994)

### DIFF
--- a/controllers/cinder_controller.go
+++ b/controllers/cinder_controller.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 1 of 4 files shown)


controllers/cinder_controller.go
```diff
@@ -357,9 +358,43 @@ func (r *CinderReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	listOps := &client.ListOptions{
+		Namespace: src.GetNamespace(),
+	}
+	err := r.Client.List(ctx, crList, listOps)
+	if err != nil {
+		l.Error(err, fmt.Sprintf("listing %s for namespace: %s", crList.GroupVersionKind().Kind, src.GetNamespace()))
+		return requests
+	}
+
+	for _, item := range crList.Items {
```
ASBishop, line 382: There was a time (long ago) when we hadn't finalized the decision that there would be just one 'Cinder' per namespace. Is there now a chance this code will need to iterate of a list of cinders, or do we expect the list will just have one entry and iteration loop is just for safety?

stuggi: yes its to keep it generic, we query for the lust of Cinder in the namespace and just loop over it, there might be none.

PR #516, opened 2025-06-24
Enable notifications for Cinder

fmount:
This patch introduces a new `notificationInstanceBus` parameter in the `Cinder` CR. This parameter is used to create a `TransportURL` that will be used for notification purposes. When a `notificationInstanceBus` parameter is set, cinder controller:

1. init the `Notification` condition provided by `lib-common`
2. in case of a dedicated `rabbitmq` instance, requests a new `TransportURL` following the known pattern
3. fetches from the resulting k8s `Secret` the info required to customize `00-config.conf`, where `oslo_messaging_notifications` is present 

Jira: https://issues.redhat.com/browse/OSPRH-17701


fmount:
After merging https://github.com/openstack-k8s-operators/manila-operator/pull/433 where there is good material that justify the approach of this patch, I think we can move forward and land this one as well.

stuggi:
> After merging [openstack-k8s-operators/manila-operator#433](https://github.com/openstack-k8s-operators/manila-operator/pull/433) where there is good material to justify the approach of this patch, I think we can move forward and land this one as well.

I was wondering if @ASBishop wanted to review it again. thats why I only added the lgtm. I am good too get it landed

fmount:
> > After merging [openstack-k8s-operators/manila-operator#433](https://github.com/openstack-k8s-operators/manila-operator/pull/433) where there is good material to justify the approach of this patch, I think we can move forward and land this one as well.
> 
> I was wondering if @ASBishop wanted to review it again. thats why I only added the lgtm. I am good too get it landed

ack, let's hear more from @ASBishop and then we can land it. In general we tend to keep manila and cinder operators in sync.
@ASBishop  to give you more context from the previous change, the tl;dr is that we had to slightly change the implementation because having `TransportURLSecret` as a `map[string]string` in `.Status` resulted in breaking the current minor updates process (more info on https://github.com/openstack-k8s-operators/manila-operator/pull/433#discussion_r2166493522)

PR #528, opened 2025-07-24
Fix panic due to missing memcachedInstance in the subCR

fmount:
We currently assign to the `subCR` spec a `memcached` instance under certain conditions.
We observe a `panic` for the `subcontrollers` (`api/scheduler/backup/volumes`) on [1] (the same applies to sched and volumes) when it tries to dereference a field that is a `nil` pointer, and this let the operator crash without any chance to recover. 
Regardless of the `memcached` CR config (`TLS` or not), we always wait in the main controller for it to be ready, and only then we unblock the rest of the reconciliation loop.
For this reason, we can always assign to the underlying components the `MemcachedInstance` defined at the top-level because:
1. we know it exists
2. we need to pass it down to the `StatefulSet` and we do not support a different `memcached` instance per component (we only support inheritance from the top-level)

Based on the above, the `subCR` parameter is not independent, and needs to always be assigned from the top-level controller. The `StatefulSet` produces the right configuration according to the `memcached` instance retrieved from the namespace.
This patch fixes the panic by always assigning the `MemcachedInstance` to the `subCR` template (a pointer) by inheriting the same value from the top-level CR.

[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/controllers/cinderapi_controller.go#L921

### DIFF
--- a/controllers/cinder_controller.go
+++ b/controllers/cinder_controller.go
@@ -773,7 +773,7 @@ func (r *CinderReconciler) reconcileNormal(ctx context.Context, instance *cinder
 	//
 
 	// deploy cinder-api
-	cinderAPI, op, err := r.apiDeploymentCreateOrUpdate(ctx, instance, memcached)
+	cinderAPI, op, err := r.apiDeploymentCreateOrUpdate(ctx, instance)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


stuggi (approved):
/lgtm

PR #529, opened 2025-07-25
Fix notificationsBusInstance typo in the API parameter

fmount:
`Nova` and probably other operators have `notificationsBus` instead of `notificationBus`, which let storage operator to slightly diverge due to the missing final `s`.
This patch fixes the interface (currently present only in main).


fmount:
There currently a quay.io outage. We'll re-kick the test once it is resolved.

fmount:
@stuggi ready to go!

PR #548, opened 2025-09-19
Refactor: Make Cinder WSGI path conditional

raukadah:
https://review.opendev.org/c/openstack/cinder/+/960831 Remove pbr's wsgi_scripts from setup.cfg and recommends to use should instead reference the Python module path for this service, cinder.wsgi.api.

Since the cinder.wsgi.api path might be different for older release that's why we want to handle the path on operator side and point the path to python module directly.

Closes: [OSPCIX-1052](https://issues.redhat.com//browse/OSPCIX-1052)

### DIFF
--- a/templates/cinder/config/10-cinder_wsgi.conf
+++ b/templates/cinder/config/10-cinder_wsgi.conf
@@ -3,16 +3,18 @@
 <VirtualHost *:8776>
   ServerName {{ $vhost.ServerName }}
 
-  ## Vhost docroot
-  DocumentRoot "/var/www/cgi-bin/cinder"
+  ## Set DocumentRoot and Directory permissions only if the path exists
+  <If "-d '/var/www/cgi-bin/cinder'">
+    ## Vhost docroot
+    DocumentRoot "/var/www/cgi-bin/cinder"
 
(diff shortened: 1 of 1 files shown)


templates/cinder/config/10-cinder_wsgi.conf
```diff
@@ -3,16 +3,18 @@
...
-
-  <Directory "/var/www/cgi-bin/cinder">
-    Options -Indexes +FollowSymLinks +MultiViews
-    AllowOverride None
-    Require all granted
-  </Directory>
+  ## Set DocumentRoot and Directory permissions only if the path exists
+  <If "-d '/var/www/cgi-bin/cinder'">
+    ## Vhost docroot
+    DocumentRoot "/var/www/cgi-bin/cinder"
```
fmount, line 9: -1:  this will fail with the following:

```
+ exec /usr/sbin/httpd -DFOREGROUND
AH00526: Syntax error on line 9 of /etc/httpd/conf.d/10-cinder_wsgi.conf:
DocumentRoot not allowed in <If> context
```

I think we need to set `DocumentRoot` anyway, outside the `If` condition.

fmount (changes requested):

PR #610, opened 2026-02-18
Probes configuration and override support

fmount:
Add `ProbeOverrides` interface and `CreateProbeSet()` function from lib-common for `unified probe management` across `Cinder` services. Enable probe customization through `CRD` overrides and remove code duplication. Updates all services (`API`, `Scheduler`, `Volume`, `Backup`) to use the new pattern with proper scheme handling and consistent defaults.
In addition, webhook validation for probes have been introduced.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/673


fmount:
recheck

fmount:
Here's a quick comparison between what we had before this patch (in terms of defaults) and what we have now (regardless they are used or not):

``` 

  ┌───────────┬─────────────────────┬─────────────────┬────────────────┬────────┐
  │   Probe   │        Field        │ Old (hardcoded) │ New (computed) │ Change │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┼────────┤
  │ Liveness  │ TimeoutSeconds      │ 5               │ 5              │ same   │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┼────────┤
  │ Liveness  │ PeriodSeconds       │ 3               │ 20             │ +17    │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┼────────┤
  │ Liveness  │ InitialDelaySeconds │ 3               │ 15             │ +12    │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┼────────┤
  │ Readiness │ TimeoutSeconds      │ --              │ 5              │ new    │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┼────────┤
  │ Readiness │ PeriodSeconds       │ --              │ 20             │ new    │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┼────────┤
  │ Readiness │ InitialDelaySeconds │ --              │ 15             │ new    │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┼────────┤
  │ Startup   │ TimeoutSeconds      │ 5               │ 5              │ same   │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┼────────┤
  │ Startup   │ PeriodSeconds       │ 5               │ 10             │ +5     │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┼────────┤
  │ Startup   │ InitialDelaySeconds │ 5               │ 20             │ +15    │
  ├───────────┼─────────────────────┼─────────────────┼────────────────┼────────┤
  │ Startup   │ FailureThreshold    │ 12              │ 12             │ same   │
  └───────────┴─────────────────────┴─────────────────┴────────────────┴────────┘
  ```

**Comment by TristanCacqueray** [comment]
recheck

**Comment by TristanCacqueray** [comment]
recheck

File: internal/cinderapi/statefuleset.go [hunk 1]
```diff
@@ -43,38 +43,33 @@ func StatefulSet(
...
+		scheme = corev1.URISchemeHTTPS
 	}
-	readinessProbe := &corev1.Probe{
-		// TODO might need tuning
-		TimeoutSeconds:      5,
-		PeriodSeconds:       5,
-		InitialDelaySeconds: 5,
+
+	// Note that by default we create probes with the same URIScheme and port
+	apiProbes, err := probes.CreateProbeSet(
```
fmount, line 57: @abays this is exactly the same patch we did in manila, can you take a look and see if we can land this as well?

api/v1beta1/cinderbackup_types.go
```diff
@@ -21,8 +21,15 @@ import (
 	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/probes"
 )
 
+// CinderBackupOverrideSpec to override the generated manifest of several child resources.
```
abays:
```suggestion
// BackupOverrideSpec to override the generated manifest of several child resources.
```

test/functional/cinder_controller_test.go
```diff
@@ -2064,6 +2064,114 @@ var _ = Describe("Cinder Webhook", func() {
 			}, timeout, interval).Should(Succeed())
 		})
 	})
+
+	When("Cinder CR instance is built with custom probes", func() {
```
abays, line 2068: Is there any particular reason `CinderBackup` isn't covered here as well?

api/v1beta1/cinderbackup_types.go
fmount: thanks, fixing this.

test/functional/cinder_controller_test.go
fmount: The reason I think is that we currently do not test `cinderBackup` in envTests, and we should! Let me try to add cbak basic coverage for `probes`, and we can keep any refactoring for follow up patches.

tl;dr we don't test `cinder-backup` at all with envTest, but I'm going to enable it as part of an additional commit.

fmount: Ok I added basic support for CinderBackup. I think it deserves a dedicated PR to go through the interfaces and make sure we are able to cover everything, but here the scope is to try probes override, so we should be good. Thanks @abays, this is a good catch and as a result of that I enabled a missing component in our envTests.

api/v1beta1/cinderbackup_types.go
fmount: Done

test/functional/cinder_controller_test.go
abays: Cool man :+1:

abays (approved):
/lgtm

PR #624, opened 2026-04-01
Add support for cinder-manage online data migrations

fmount:
Enhance the existing job interface to support multiple `cinder-manage` commands.
The main purpose of this patch is to introduce the ability to run `online_data_migration`, required in the upgrade context and described in the upstream documentation [1].
For this reason, this patch:
- adds a generic `CinderManageJob()` function with job type parameter
- implements an `OnlineDataMigrationsJob()`, useful for upgrade scenarios
- keeps backward compatibility with existing `DbSyncJob()`

Jira: [OSPRH-1559](https://redhat.atlassian.net/browse/OSPRH-1559) 

[1] https://docs.openstack.org/cinder/latest/admin/upgrades.html

### DIFF
--- a/internal/cinder/job.go
+++ b/internal/cinder/job.go
@@ -8,6 +8,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// ManageJobType defines the type of cinder-manage command to run
+type ManageJobType string
+
 const (
 	// DBSyncCommand -
 	// TODO: Once we work on update/upgrades revisit the command in the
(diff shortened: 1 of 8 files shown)


fmount:
I rebased this patch and I'm wondering if @abays || @stuggi can take a look at this.
The idea is to run an additional `cinder-manage` command (`online-data-migration`) that won't do anything in a greenfield scenario, but it will perform the db migration in an upgrade context (see [1]).
By adding this job we don't have to provide a dedicated "upgrade path", and we can keep the original flow because the command is idempotent.

Do you have any concern about adding this step?

[1] https://docs.openstack.org/cinder/latest/admin/upgrades.html#database-upgrades

test/functional/cinder_test_data.go
```diff
@@ -100,6 +101,10 @@ func GetCinderTestData(cinderName types.NamespacedName) CinderTestData {
 			Namespace: cinderName.Namespace,
 			Name:      fmt.Sprintf("%s-db-sync", cinderName.Name),
 		},
+		CinderOnlineDataMigration: types.NamespacedName{
+			Namespace: cinderName.Namespace,
+			Name:      fmt.Sprintf("%s-online-data-migrations", cinderName.Name),
```
abays, line 106: Why is plural used in the name here when we're using singular everywhere else?  Not a blocker by any means, just curious if I'm missing something.

internal/cinder/job.go
```diff
@@ -51,38 +64,28 @@ func DbSyncJob(instance *cinderv1beta1.Cinder, labels map[string]string, annotat
...
 
 	// add CA cert if defined
 	if instance.Spec.CinderAPI.TLS.CaBundleSecretName != "" {
-		dbSyncVolume = append(dbSyncVolume, instance.Spec.CinderAPI.TLS.CreateVolume())
-		dbSyncMounts = append(dbSyncMounts, instance.Spec.CinderAPI.TLS.CreateVolumeMounts(nil)...)
+		jobVolume = append(jobVolume, instance.Spec.CinderAPI.TLS.CreateVolume())
+		jobMounts = append(jobMounts, instance.Spec.CinderAPI.TLS.CreateVolumeMounts(nil)...)
 	}
 
-	dbSyncExtraMounts := []cinderv1beta1.CinderExtraVolMounts{}
```
abays, line 74: This was formerly needed because of the shared `GetVolumeMounts` and `GetVolumes` functions, right?  It wasn't because db syncs ever attached extra mounts?  I'm just wondering if there could be a scenario where a previous version had extra mounts here and then for some reason db sync would run again and be missing the extra data.

test/functional/cinder_test_data.go
fmount: I think the plural is to reflect the original command `cinder-manage db online_data_migrations`, and because I originally called it `online-data-migrations` in job.go, I had to align the `envTest` accordingly. I could move it back to singular, but the original idea was to convey in the name the equivalent command run behind the scenes.

internal/cinder/job.go
fmount: Correct, and because `dbsync` and any variation of `cinder-manage` command usually needs only db connection, I simplified this part to only get what's rendered in /etc/cinder/cinder.conf.d, with no kolla in the middle (we don't need it). `ExtraMounts` was part of that interface but it's not supposed to be used (hence the empty struct instance).

abays:
+1 from me.  Will defer approval to let @stuggi have a look as well

stuggi (approved):
/lgtm I think we need something similar across service operators for upgrades. good to  have it as a reference on how we could do it.

fmount:
Thanks both for the review!!

PR #631, opened 2026-04-17
Add glance section required by Cinder to use new glance location-api

fmount:
We need to register `glance` user in `keystoneauth` plugin to allow service to service communication and let `Cinder` access to the location-api glance calls. [1].

Jira: [OSPRH-28617](https://redhat.atlassian.net/browse/OSPRH-28617)

[1] https://review.opendev.org/c/openstack/cinder/+/958716

### DIFF
--- a/templates/cinder/config/00-global-defaults.conf
+++ b/templates/cinder/config/00-global-defaults.conf
@@ -122,6 +122,25 @@ project_domain_name = Default
 region_name = {{ .Region }}
 {{ end -}}
 
+[glance]
+interface = internal
+auth_url = {{ .KeystoneInternalURL }}
+{{ if (index . "ApplicationCredentialID") -}}
+auth_type = v3applicationcredential
+application_credential_id = {{ .ApplicationCredentialID }}
+application_credential_secret = {{ .ApplicationCredentialSecret }}
+{{ else -}}
+auth_type = password
+username = {{ .ServiceUser }}
+password = {{ .ServicePassword }}
+user_domain_name = Default
+project_name = service
+project_domain_name = Default
+{{ end -}}
+{{ if (index . "Region") -}}
+region_name = {{ .Region }}
+{{ end -}}
+
 [service_user]
 send_service_user_token = True
 auth_url = {{ .KeystoneInternalURL }}


templates/cinder/config/00-global-defaults.conf
```diff
@@ -122,6 +122,25 @@ project_domain_name = Default
 region_name = {{ .Region }}
 {{ end -}}
 
+[glance]
```
fmount, line 125: @rajathere @konan-abhi does this block eventually makes sense based on [1]?

[1] https://review.opendev.org/c/openstack/devstack/+/958718
[2] https://review.opendev.org/c/openstack/cinder/+/958716
[3] https://docs.openstack.org/cinder/latest/_static/cinder.conf.sample

templates/cinder/config/00-global-defaults.conf
```diff
@@ -122,6 +122,25 @@ project_domain_name = Default
...
+interface = internal
+auth_url = {{ .KeystoneInternalURL }}
+{{ if (index . "ApplicationCredentialID") -}}
+auth_type = v3applicationcredential
```
konan-abhi, line 129: I think we need to test service to service communication in case of application credentials. In glance we explicitly check whether service role is present in context or not.

https://github.com/openstack/glance/blob/master/glance/api/v2/images.py#L1161

fmount: @Deydra71 do you know if in this case appcred works as expected? not sure if someone from keystone can verify this.

Deydra71: @fmount From the Keystone/application credential perspective, this should work as long as the application credential includes the `service` role (as I read is requred in Glance policy). In our ctlplane flow, service user application credentials are created with both `admin` and `service` roles enabled by default [1]. And in our downstream doc (yet to be released) is stated that `service` role must be enabled at all times for all service users app cred.

I am not that much familiar with the exact Glance/Cinder service to service path, so I would still validate it with an actual Cinder <-> Glance flow. But app cred shouldn't be a blocker here.

[1] https://github.com/openstack-k8s-operators/dev-docs/blob/main/application_credentials.md#global-ac-settings

konan-abhi:
Looks good to me!!

templates/cinder/config/00-global-defaults.conf
fmount: Thanks @Deydra71, then I think we can move forward with this!

stuggi (approved):
/lgtm

fmount:
just rebased.

fmount:
recheck

fmount:
the new section is properly applied [1]

[1] https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/logs//1df/rdoproject.org/1dfe5f9ec9fd4d01ba9a6c2717e2c04a/controller/ci-framework-data/logs/openstack-must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-70ce4ab587da14dddf23f3816879343650007e91038f935686ab5355e6f2a7a2/namespaces/openstack/secrets/cinder/cinder-volume-volume1-config-data.yaml-00-global-defaults.conf

fmount:
ok tests look good, @stuggi to re-approve (thanks a lot for the help!)

stuggi (approved):
/lgmt

rajathere:
Sorry for being late but this is a exact replica of the [nova] section which is desired.
We use the service user (cinder here) to communicate with glance and the only difference between upstream implementation and here is, based on wheather app creds are enabled, we make the decision to do password based auth or use app creds secret.
Overall, LGTM.

PR #635, opened 2026-04-21
watch to cinder subCR secrets

fmount:
When a `config-data` `Secret` owned by a `cinder` sub-controller (`API`, `Scheduler`, `Backup`, `Volume`) is manually deleted, it is never recreated because none of the sub-controllers watches for changes to `owned` `Secrets`. Add `Owns(&corev1.Secret{})` to all four sub-controller `SetupWithManager` functions, matching the pattern already used by glance-operator, barbican-operator, neutron-operator, etc.
This ensures that deleting an owned secret triggers reconciliation, which recreates it via `.EnsureSecrets()`.

Jira: [OSPRH-29330](https://redhat.atlassian.net/browse/OSPRH-29330)


fmount:
recheck

stuggi (approved):
/lgtm

